PR #2, opened 2019-04-14
Updating to specify dependencies.

augustjohnson:
Hardcoding the dependencies which were causing an issues related to "allow_unicode" and slugs in the search indexes.


augustjohnson:
That looks like it fixed it.

PR #5, opened 2019-05-27
Updating Readme and Creating a /weapons endpoint.

augustjohnson:


augustjohnson:
Heyo, this should be an easy merge

eepMoody:
Looking into it right now! It looks like it isn't building in dockercloud for some reason. Going to re-run, but if it doesn't work could you try merging in from Master just to be safe?

augustjohnson:
I haven't checked your dockercloud stuff... but I did modify the weapons doc and I added a /weapons endpoint.

PR #9, opened 2019-07-17
Add Weapons json data

gbalbuena:


augustjohnson:
I could probably clean this up and make a /weapons endpoint...  Does that
jive with what you were thinking?

gbalbuena:
I need the data, this is actually a `json` version of a `yaml` in my personal project, please be free to use it or change at will.

Don't have time to learn django, but a pr fresh with the code will give me an idea for future contributions

eepMoody:
Yeah this is super handy! I think before we build an endpoint off it I may want to try and remap a few things to be more consistent with our existing data but the data itself is an awesome starting point! 

eg. attacks from monsters have 'damage_dice', while these have the damage & type together. But this is all so well-structured already that should be super simple and something I can bust through this weekend with a regex or something!

PR #21, opened 2019-09-20
Add Creature Codex stuff.

RealDeuce:
This is likely broken... and I'm just guessing at tables and line breaks.


RealDeuce:
I suspect how I shoe-horned the OGL into the file is a part of why the check failed... but I can't actually see the details.

eepMoody:
@RealDeuce it looks like the only issue was that your actions didn't include an "attack_bonus" except where it actually makes sense. The other files we used always included it, but it was simple enough to make the importer more flexible so we don't have to worry about it!

Assuming this passes, I think this PR is good to go.

RealDeuce:
Did the table in the Elophar "Possessed by Ancestors" ability work?

eepMoody:
@RealDeuce Took me a bit to find the right way to test this, but yeah the tables look good! Gonna merge this in.

PR #23, opened 2019-10-02
'Light' (weapon property) should be a subheading

michaeleconomy:


eepMoody:
Nice catch! Thanks @michaeleconomy

PR #30, opened 2019-10-24
Windows machine encoding bugfix

colbyrush:
Noticed encoding issues for the database, preventing indexing of everything except first two spell items, on a windows machine. Resolved by forcing 'encoding="latin-1"' for each file section in api/management/commands/populateddb.py


colbyrush:
Changed latin-1 to utf-8 for more universal usage

PR #33, opened 2019-10-28
Add Armor json data #9

gbalbuena:
This is the armor json file that I'm using to calculate ac, was thinking could be valuable with some corrections.

### DIFF
--- a/data/WOTC_5e_SRD_v5.1/armor.json
+++ b/data/WOTC_5e_SRD_v5.1/armor.json
@@ -103,7 +103,7 @@
       "ac": 14,
       "cost": "30 gp",
       "weight": "40 lb.",
-      "modifier": [ "dexterity" ],
+      "modifier": [ ],
       "requirements": {},
       "disadvantage": true
     }, {
@@ -114,7 +114,7 @@
       "ac": 16,
       "cost": "75 gp",
       "weight": "55 lb.",
-      "modifier": [ "dexterity" ],
+      "modifier": [ ],
       "requirements": { "strength": 13 },
       "disadvantage": true
     }, {
@@ -125,7 +125,7 @@
       "ac": 17,
       "cost": "200 gp",
       "weight": "60 lb.",
-      "modifier": [ "dexterity" ],
+      "modifier": [ ],
       "requirements": { "strength": 15 },
       "disadvantage": true
     }, {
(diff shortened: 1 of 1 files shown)


data/WOTC_5e_SRD_v5.1/armor.json
```diff
@@ -0,0 +1,187 @@
...
+      "disadvantage": true
+    }, {
+      "id": "breastplate",
+      "name": "Breastplate",
+      "type": "Medium Armor",
+      "rarity": "Standard",
+      "ac": 14,
+      "cost": "400 gp",
+      "weight": "20 lb.",
+      "modifier": [ "dexterity" ],
```
eepMoody: On these guys I think this is supposed to be dex (max 2). Is there any way to indicate that here?

data/WOTC_5e_SRD_v5.1/armor.json
```diff
@@ -0,0 +1,187 @@
...
+    }, {
+      "id": "plate_mail",
+      "name": "Plate Mail",
+      "type": "Heavy Armor",
+      "rarity": "Standard",
+      "ac": 18,
+      "cost": "1500 gp",
+      "weight": "65 lb.",
+      "requirements": { "strength": 15 },
+      "modifier": [ "dexterity" ],
```
eepMoody: Similar to the medium armor, shouldn't this be blank? According to the rules Plate doesn't add any bonus.

gbalbuena: good catch

data/WOTC_5e_SRD_v5.1/armor.json
gbalbuena: what do you think about about add `"modifier_max": 2`

gbalbuena: This is the code to calculate ac using that data source

https://github.com/gbalbuena/dndcli/blob/master/src/core/index.js#L3-L43

gbalbuena:
> Hey guys, just so you know, I took this general file and after building a model and import process that would work, I adjusted it (quite a bit actually). You can see my work over in PR #38. Technically that work completely supercedes this PR, but it absolutely was based on this file.

Good work!, we haven't finalize the structure, I'm about to send and update around the structure of the data based on @eepMoody comments

gbalbuena:
please let me know if any change is required to merge this, so far I'm using it for my ac formulas and It's working as expected, you might need to do changes for your requirements or keep it as it is, we can always change and adapt

gbalbuena:
@eepMoody Legendary! ⚔️

Issue #36, opened 2019-10-28
Django Rest Framework AUTH is enabled and unused.


augustjohnson:
We should remove as it's an attack surface.
https://api.open5e.com/api-auth/login/?next=/

Issue #34, opened 2019-10-28
Django debug mode is on in production


augustjohnson:
See here.
https://api.open5e.com/search/

augustjohnson:
I believe this SHOULD be addressed (depending on the content of the .env file in production) by PR #37

PR #37, opened 2019-10-28
Removing authentication controls and permissions. #36

augustjohnson:
Fixes issues #36

### DIFF
--- a/server/settings.py
+++ b/server/settings.py
@@ -27,7 +27,7 @@ assert 'SECRET_KEY' in os.environ, 'Set SECRET_KEY in your .env file!'
 SECRET_KEY = os.environ['SECRET_KEY']
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.environ.get('OPEN5E_DEBUG', False)
+DEBUG = os.environ.get('OPEN5E_DEBUG', '') != 'False'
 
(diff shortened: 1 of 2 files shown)


augustjohnson:
Well that teaches me to do my dev on my local branch.  This ALSO includes some DEBUG work related to issue #34

eepMoody:
@augustjohnson Sorry, I'm not really following what this does. Can you give a quick explanation just so I have something to refer back to if it ever comes up?

augustjohnson:
On this page (https://api.open5e.com/) there's a Log in link in the upper right.  That link has no reason to exist.  The site offers no ability to auth, and all functions are available unauthenticated.  This PR is to remove the potential ability to log in at all (as it could be abused potentially).  

The benefit of this PR is a lower attack surface. There is no login form, anything posted to it will now be ignored, and all requests are now explicitly "Anonymous," with 0 authorization. There is no change in functionality other than an enhanced security posture.

augustjohnson:
Ah, and the last benefit is that this actually will read the word "false" from .env and set the debug value to False, the python boolean value, not "false" the string (so it will no longer use debug mode).

augustjohnson:
Thanks for the merge!  So I have identified that your .env in production does not include OPEN5E_DEBUG=False, so debug is still on in production...  I think this is a manual change somewhere.  it needs to be that exact string.

PR #38, opened 2019-10-29
Armor endpoint

augustjohnson:
Inspired by @gbalbuena, I modified his data, and added the /armor endpoint, along with the SRD data.  This includes a couple of non-armor armors, but should be OK.  Source data includes "rarity" which is not imported.


eepMoody:
I'm not really sure how best to resolve this against @gbalbuena 's MR. @augustjohnson What do you two think? Should we close the initial MR down and just do this one, or is there some benefit to trying to merge/rebase these two branches?

augustjohnson:
I think you could merge his then merge mine (overwriting the armor.json file).  That would give him history, and still work.

gbalbuena:
as long as It's expressive for the purpose of the usage I'm ok with the merge data approach and adapt to site needs, the implementation I did was with a backend purpose not a public api for web

eepMoody:
@augustjohnson @gbalbuena  So I know this one has sat for a while, I'm just not sure what to do with it. According to the diff every line in the JSON file is a merge conflict.

Would it work if we just excluded the armor.json update? Or should we go ahead and take this one?

PR #41, opened 2019-11-06
Lint JSON files

txtsd:
Lint all json files so that programmatic editing won't cause unnecessary changes in whitespace and indentation.


txtsd:
Bork'd my branch. I'll redo it.

PR #58, opened 2020-05-01
Fix incorrect formatting in "Adventuring Gear"

cparadis6191:
Add missing newlines to "Crowbar" and "Healer's Kit".
Add missing emphasis on "Healer's Kit".


cparadis6191:
`git diff --word-diff HEAD~` gives a much better looking diff.

PR #81, opened 2020-12-09
Update monsters.json

michaeleconomy:
thats a 1 not an L


eepMoody:
Hm, not sure why this shows waiting on dockercloud tests, but it looks like it built just fine! I'll get it merged in.

PR #117, opened 2023-03-09
Address installation issues

Gredelston:
I hit a few installation issues while getting started with this project. To prevent those issues for future devs, I've added a setuptools dependency, and added some prerequisite notes to the README.

Note: I accidentally made these commits from my @google.com account instead of my @gmail.com account. I don't think that should matter, but I will do better to not make that mistake in the future. >_<


Gredelston:
@eepMoody FYI

README.md
```diff
@@ -37,14 +53,15 @@ At that point, you will be able to run the django server normally (within the pi
...
+
 This deployment has been tested using DigitalOcean with Docker.
 
 To start up the server from scratch on a droplet:
 
 ```
-git pull https://github.com/eepMoody/open5e-api
+git pull https://github.com/open5e/open5e-api
 export SECRET_KEY=a_new_secret_key
 export SERVER_NAME=whatever.open5e.com
```
eepMoody, line 64:
```suggestion
export SERVER_NAME=whatever.yourdomain.com
```

A small change (since most consumers won't be using open5e.com) that has the added benefit of kicking off a new CI/CD build! Sorry about that disruption, switching to an org had some unintended side effects.

Issue #84, opened 2021-03-02
Ordering by CR descending puts creatures with CR 9 at the top.


tonesija:
Seems to me it doesn't sort the CR value like a number but like a string so it returns that CR of 9 is above CR 21.
Example: hitting this endpoint: https://api.open5e.com/monsters/?ordering=-challenge_rating&page=1 I get that the first monster is Clay Golem which has a CR of 9.

paullofte:
I'm not sure how to get this into the code base but I found this works for converting the string CR to a number that can be properly sorted

from fractions import Fraction
cr_str = '1/4'
cr_f = Fraction(cr_str)
cr = float(cr_f)
0.25

Gredelston:
Nice fix, @paullofte. Looks like this works now.

API view with CR descending, I see Tarrasque at the top, as expected: https://api.open5e.com/monsters/?ordering=-cr&page=1
API view with CR ascending, I see Awakened Shrub at the top: https://api.open5e.com/monsters/?ordering=cr&page=1

Of course, the `challenge_rating` field is still stringy and thus will sort wrong. Perhaps we should remove that field to avoid confusion.

eepMoody:
I'm hesitant to remove a field people may be consuming, which I'm betting
that one is. I think maybe it would be best to clarify in our documentation
that that's the text "display" version of the value (as provided in the
SRDs) and that "cr" acts as the numerical/programmatic field?

augustjohnson:
What about adding the float CR as a “cr_float” new field? Or “cr_numerical” or something. That would totally maintain backward compatibility.

augustjohnson:
Ok, @Gredelston's fix provides a new field (cr), which is a orderable floatfield. He properly exposes it, and showed that it works.  This is good to go.

PR #121, opened 2023-03-10
Fixes issue #84, add numeric CR for sorting and range based operations

paullofte:
Adding a second CR field to monsters that is the numeric equivalent so that the challenge_rating can be displayed as the friendly 1/8 and 1/4 style but sorting can be done by CR


eepMoody:
I really like this solution. We should think about if there's any other places we need to deal with fractions and repeat this pattern!

PR #122, opened 2023-03-10
Clean up populatedb.py and importer.py

Gredelston:
This PR contains significant changes to `api/management/commands/populatedb.py` and `api/management/commands/importer.py`. The changes generally remove duplicated code and hopefully make it easier to add new weird types of stuff in the future.
I've also added docstrings, a few new helper classes, stuff like that, which I hope will help future devs understand what we're doing here.

This change is quite invasive, so it warranted careful testing. To that end, I ran `quicksetup` on HEAD and dumped the database to a json file, and then did the same on this branch. The only differences were timestamps, primary keys, and the fact that Armor doesn't get Manifests on ToT, which I assume was a bug that I have conveniently fixed.

There are also a few random import changes in this PR: imports are now grouped in a standardized fashion compliant with PEP-8, and I have removed all wildcard imports (also in compliance with PEP-8).


Gredelston:
@eepMoody FYI! (I may have neglected my job for the last two days because I was having fun coding.)

Gredelston:
Lmk what you think of the new pattern here. Happy to make edits. My hope here is that by reducing duplicated code we make it easier to add new code and harder to introduce bugs.

eepMoody:
Holy crap my man, this looks incredible! Went through this and everything else today (I have also been neglecting my job a bit out of renewed excitement!)

Merging this and your other changes in.

Gredelston:
Thanks! Not sure if you meant to merge this but it is still open.

augustjohnson:
FYI I’ve done a cursory phone-based review and I like the changes broadly but I couldn’t  test. I’m definitely feeling some gratitude for you building the foundation for more contributions.

eepMoody:
Doing some merging with other stuff ongoing, apologies for the delay.

PR #130, opened 2023-03-14
cant -> can't

BramvdnHeuvel:
Cannot is abbreviated as can't, not cant. This pull request fixes that spelling mistake in one document where this mistake happens quite often.

NOTE: It's more than a simple find-and-replace-all to prevent Thieves cant from being changed to Thieves can't, which would be incorrect.


eepMoody:
Nice catch, thank you!

PR #132, opened 2023-03-14
wotc: Add hit dice modifiers for beasts and NPCs

### LINKED ISSUE #100: Many statblocks in `monsters.json` are missing the modifier for their `hit_dice` field
https://github.com/eepMoody/open5e-api/blob/master/data/WOTC_5e_SRD_v5.1/monsters.json
A few examples include the Bandit and Knight statblocks, but any instance of the `hit_dice` field where there is no modifier is probably affected by this.

Improved Initiative uses the hit_dice field when rolling random monster HP so the modifier is important.
https://github.com/cynicaloptimist/improved-initiative/issues/580

Thanks!

Gredelston:
Fixes #100. For some reason, the WotC monsters.json was missing modifiers (ex. the "+2" part of "2d8+2") for all beasts (ex. Wolf) and NPCs (ex. Bandit). I've manually flipped through the WotC PDF and corrected the "hit dice" field for all monsters (to the best of my human ability).

I've also found and corrected a few HP data errors for other (non-beast, non-NPC) monsters.

Tested via `manage.py populatedb --update data/WOTC_5e_SRD_v5.1/`.


eepMoody (approved):
Nice!

Issue #141, opened 2023-03-18
API should expose page numbers


Gredelston:
We have received a request to populate monsters' page numbers, where possible, and expose them on the API. In particular, we have received page# data for `data/menagerie/monsters.json` from Mike Shea. In order to support this FR, we'll need to:

1. Create a new page# field. I propose that we do this on the `GameContent` abstract model, which will make it available on the `Monster` table as well as on other tables such as `Spell`, `CharClass`, etc. I propose that we call the field `page_no`. When a data source has a physical version, `page_no` should reference the physical page number, even if the PDF version has extra pages at the beginning (ex. cover page). If the PDF version is the canonical source, then that's moot.
2. Populate the field in `data/menagerie/monsters.json` via the data Mike provided.
3. Expose the new field on the Monsters API view.

Gredelston:
Here's the Monstrous Menagerie page#s, courtesy of both Mike and Ravine: [monster_index_02.md](https://github.com/open5e/open5e-api/files/11009211/monster_index_02.md)

Notes, from Ravine:

```statblock exists, monster does not exists in the book
- behir magus (does not exists in book)
- horde of shadows (does not exist: squad template applied to shadows)
- miremuck goblin king (no idea)
- nilbog (no idea, shows as from "Other" in a5e.tools)
- thunderbird (no idea)
- yobbo ( ¯\_(ツ)_/¯ )

missing statblock (variant or entry exists in book, does not have a block in the data)
- adamanticore (manticore variant)
- flame-spitting statue, 24
- cannibal reaver, 247
- chained one, 82
- firebird, 504
- lacedon ghast, 229
- lacedon ghoul, 229 (called "lacedon", looks like the ghoul block with lacedon swimming trait)

I removed all the chimera entries as they do not exist as variant statblocks on the website as far as I can tell; they are also minimal changes.
- faerie chimera, 46
- fiendish chimera, 46
- sea chimera, 46
- sewer chimera, 46
- tundra chimera, 46

Some entries do not exist in the index but exist in the source data (added)
- naiad
- stegosaurus
```

Gredelston:
And here's the Tome of Beasts 3 data as JSON
[tob3_page_numbers.txt](https://github.com/open5e/open5e-api/files/11009215/tob3_page_numbers.txt)
, courtesy of Mike.

PR #149, opened 2023-03-19
Add monster page # (where data is available)

Gredelston:
With this pull request:

1. We add a new field `page_no` to all descendants of `GameContent`.
2. We expose that new field on the Monster serializer and viewset, since that's the only table for which we have any page_no data yet.
3. We populate `page_no` for the monsters in the Monstrous Menagerie, Creature Codex, and Tomes of Beasts #1–3 (data provided by Mike Shea).

Tested by running `quicksetup` to populate and then checking `localhost:8000/monsters` to verify that the page numbers are exposed on the API.


Gredelston:
Fixes #141

eepMoody (approved):

PR #152, opened 2023-03-19
Adding page numbers based on the SRD PDF

augustjohnson:


eepMoody (approved):

augustjohnson:
Hey there's some weird unicode characters in this... take a triple check before merging. I'm not sure I MEANT to commit those, will those cause problems?

eepMoody:
@augustjohnson Yeah, I _think_ those will be fine, but for the sake of consistency would you mind trying to find and replace them back to the actual bullet character?

Like I think our setup will digest them just fine, but other people may not be parsing them out.

eepMoody (changes requested):
Can we switch these bullets back, just for the sake of consistency? I'm not sure our build will be affected, but others might be expecting them, especially if they're doing any parsing of the spellcaster info.

augustjohnson:
Yep, should be fine now.

eepMoody (approved):

PR #155, opened 2023-03-19
populatedb: Wrap each data dir in an atomic txn

### LINKED ISSUE #143: Bulk create/update models during populatedb
Right now, `quickload` and `quicksetup` take a long time to run. It would be nice to make them faster, especially since we'll probably add more content sources in the future.

The primary reason for the slowness is saving models to the DB. Currently, `importer.py` saves each model one-at-a-time. Django provides `bulk_create()` and` bulk_update()` methods for saving a whole bunch of models at once. We should start using those. I think it could cut our `populatedb` time by 50% or more.

Gredelston:
On my laptop, this speeds up `quickload` by 98.5%, from 12.25 minutes down to 11 seconds. Pretty good, eh?

Fixes #143

### DIFF
--- a/api/management/commands/populatedb.py
+++ b/api/management/commands/populatedb.py
@@ -21,6 +21,7 @@ from pathlib import Path
 
 import django.apps
 from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
 
 from api.management.commands.importer import Importer, ImportOptions, ImportSpec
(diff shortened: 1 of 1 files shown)


eepMoody (approved):

eepMoody:
Damn this is impressive. Very cool.

PR #156, opened 2023-03-19
Using DigitalOcean App Platform

dprothero:
This is an experiment to test using DigitalOcean's App Platform. Using that platform is more expensive ($20 - $25/mo for 2GB -- assuming we need 2GB, instead of $12/mo for a 2GB droplet). That's the con. Here are the pros:

- We are no longer responsible for the underlying host operating system (which is 3 years unpatched, from what I understand)
- There will never be a need to SSH into a server to troubleshoot something.
- We can use a true SSL certificate, but don't have to manage it.
- No infrastructure automation would need to be built to manage the vm host OS.
- The solution has less moving parts and fewer things to troubleshoot (making it more accessible to contributors?)

I'm not advocating for spending more money for money's sake, but it does have some compelling advantages. It's also not a one-way door. You could easily enough go back to hosting on a VM instance.

## Testing

I've deployed an example of this to my personal DO account and you can test it here (assuming I haven't shut down the app, ping me in Discord if it's down):

https://seal-app-7vc2c.ondigitalocean.app/

## About the Dockerfile changes

Pretty simple, really. I removed the generation of a self-signed cert and just have gunicorn running normal HTTP on port 8888. The way DigitalOcean Apps work is this port isn't accessible to the Internet. It proxies for you, along with a valid, signed cert.

The cool thing is, this is the way other hosting providers like AWS AppRunner do it, so the solution is portable.

The other change is the removal of the `.env` file. You configure the environment variables for the app in the DigitalOcean control panel (and it can keep the SECRET_KEY encrypted).

## AppRunner from AWS

AWS AppRunner's pricing is pretty much the same. It's more complicated to calculate, but you end up with $25/mo. I would say it's not worth switching.

## What's next?

If we want to move forward, here's what needs to happen:

- [X] 1. Merge/close any pending open PRs
- [X] 2. Create a new branch `main` off of `master`
- [ ] 3. Change the default branch of this repo to `main`
- [X] 4. Reopen this PR as wanting to merge into `main`.
- [x] 5. Merge this PR.
- [ ] 6. Build a docker image and push it to a different tag (e.g. `docker.io/eepmoody/open5e-api:app-image` - this has already been configured in DockerHub)
- [ ] 7. Create a new App in DigitalOcean (screenshots shown below) and configure the env vars
- [ ] 8. Test on the DO App URL you are given.
- [ ] 9. Change CloudFlare to access the DO App URL (don't use IP, as this circumvents DOs magic hosting sauce)
- [ ] 10. Shut down your droplet
- [ ] 11. Monitor

<details>
<summary>Open for DigitalOcean detailed setup screenshots</summary>

1. First, go to Apps: https://cloud.digitalocean.com/apps
2. Click "Create App"
    image
3. Choose "Docker Hub" for the source
    image
4. Input the repository name and a tag name and click "Next"
    image
5. Click "Edit Plan" to choose the right plan and number of instances (it selects the Pro plan + 2 instances by default)
    image
6. Click the "< Back" button when done updating the plan
    image
7. Click "Next"
11. Click to "Edit" your environment variables for the container (not the global ones, though those probably would work)
    image
12. Add your environment variables. Use the "Encrypt" option for the `SECRET_KEY` variable. For `SERVER_NAME`, just put a placeholder value for now (since we're creating a new app, we don't know the server name yet).
13. "Save" the environment variables and click "Next"
14. You should be here, where you can click "Next"
    image
15. You then get a review page, where you can then click "Create Resources" to continue.
16. It should finish deploying (the first one will fail because we need to change the port)
17. On the "Settings" tab, click the component for your docker container:
    image
18. Scroll down to "HTTP Port" and click "Edit"
    image
19. Change the port to `8888` and save that change
20. That will kick off another deployment, at which point, we'll get our URL to our site
    image
21. On the "Settings" tab again, make sure the component for the docker container is clicked, and then scroll down to "Environment Variables" and click "Edit"
    image
22. Edit the `SERVER_NAME` variable and put the new host name for your app there.
23. This will kick off another deployment and everything should be working now!
</details>

Clean-up:

1. Now you can resume just pushing to the `latest` tag.
2. It's easy to update the settings in the DO App to use the `latest` tag instead of `app-image`

Future deployments:

1. Build/push the docker image
2. Go into the DO control panel and initiate a new deployment.

## Future improvements

The deployment can (and probably should) be automated. You could trigger a GitHub Action to run whenever you create a "Release" in GitHub (there's really no artifact to post, but the Release is still a good record of a deployment). The GitHub action could build the docker image, push it up, and then trigger the DO App deploy via API. These will require API keys, but GitHub has the ability to configure secrets in a repository for the Actions to make use of.


Gredelston:
Nifty! Is the AppPlatform machine comparable to what we have today?

dprothero:
> Nifty! Is the AppPlatform machine comparable to what we have today?

Yes, same configs are available.

eepMoody:
So I notice that step 1 is "merge to master". As the pipeline is currently set up, that will automatically deploy it over to the live DO droplet, which I suspect will cause it to go down due to the lack of that .env file and the signed certs.

To that end, I'd suggest one change: Make sure we've closed all open PRs, then make a new `main` branch that we will begin using in place of "master", which is a change I've been meaning to make anyways. 

We can have Docker Hub start building our new image from _that_ branch, and won't need to fiddle with anything afterwards.

Then we can follow the rest of the steps, switch everything over to the new server in Cloudflare, and shut down the old ones once that's complete. Master becomes a historical branch at that point.

Does that seem like a good plan?

dprothero:
@eepMoody Sounds good. Won't Watchtower see the new image pushed in DockerHub, though, and pull it down?

eepMoody:
@dprothero it's pinned to `open5e-api:latest` in the watchtower config, which is built from `master` branch. So I _think_ we'll be good?

image

eepMoody:
Sorry, I realize I was unclear about one detail: 

We close every open PR _except this one_, branch `main` off of `master`, then merge this to `main` as the first commit to the new main branch of the project.

dprothero:
@eepMoody What tag will DockerHub use when building an image off `main`? I was assuming it would also use `latest`, but I could be wrong.

eepMoody:
@dprothero I was thinking we do as you suggested, and set it to `:app-image` while we're doing the swapover, then switch `:latest` to build off `main` too once everything is stable?

eepMoody:
I went ahead and configured it in advance, even. As soon as `main` exists, it will start deploying to `app-image` automatically!

PR #159, opened 2023-03-20
A5e Feats

Sturlen:
Added content in a new data folder. The model changed so I've modified the serializer and done a migration.

Tested that I can access the /feats endpoint and I get the expected data while running the server.

Added the command to load the data in the Dockerfile, but I haven't been able to test it myself.

Didn't see any documentation on adding sources, so I just went by what earlier PR's did. Let me know if there are any problems


eepMoody (approved):
Nice, addition, and thank you for all the discussion around the best way to tackle this!

PR #170, opened 2023-03-23
Various Kobold Press OGL Spells

Sturlen:
Spells that were missing from the API, published by Kobold Press and available under the OGL. Scraped from the community made Foundry plugin at: https://github.com/dimasfh1243/fdrg/blob/main/Foundry%20VTT%20Modules%20for%20DnD%205E/Kobold%20Press%20OGL.md

Currently it's only the content file. Feel free to check if anything is missing. 

Known issues: Spell school and class were not present in the dataset.


Gredelston (approved):
Very nice work, thank you for getting this data!!

somanyrobots:
If these were translated from a foundry module, @Sturlen do you have the scripts available you used to convert them? I have several hundred more spells I'd like to PR up via the same method

Sturlen:
@somanyrobots it's been a few years since I worked on this, and I'm afraid I no longer have the scripts. Also, I'm not sure what the output should look like anymore after the move to the V2 API.

PR #178, opened 2023-03-25
Kobold Press Spells Implementation

Sturlen:
Will make the new spells available through the API.

I'm a bit unsure about the license information in `document.json`, as these spells are from a Foundry Module which includes spells from multiple different books. I've included the full contents of their license just in to be on the safe side.

Tested with the updated dependencies and the content shows up correctly in the request. Filtering by the document slug works too.

Not tested with Docker,


data/kobold_press/document.json
```diff
@@ -0,0 +1,245 @@
+[
+    {
+        "title": "Kobold Press OGL Content",
```
eepMoody, line 3: Is there any chance we can get more specific here, since there's a lot of Kobold Press OGL Content? Maybe we call it "Kobold Press OGL Spells Compilation" or similar?

data/kobold_press/document.json
```diff
@@ -0,0 +1,245 @@
...
+        "slug": "kp",
+        "desc": "Kobold Press Community Use Policy",
+        "license": "Open Gaming License",
+        "author": "Various",
+        "organization": "Kobold Press",
+        "version": "1.0",
+        "copyright": "© Open Design LLC",
```
eepMoody, line 10: Shouldn't this copyright go to Kobold Press?

eepMoody (changes requested):

dprothero:
Does the source data have what publications these are from? Or, maybe we could figure it out by cross-referencing their spell publications? Having more spells is great, but I wonder if folks would prefer to know what books things are from.

data/kobold_press/document.json
Sturlen: image
Open Design LLC seems to be the owner of the Kobold Press name. See footer: https://koboldpress.com

Sturlen:
> Does the source data have what publications these are from?

Each spell has a list of abbreviated source names like this: `HH,DM:Ru`, so it could be possible to split them into separate documents based on their source. I'm not sure how to split spells that have multiple sources though.

augustjohnson:
Hey Sturlen, I think the ideal situation is that each of those (abbreviated source names) refers to it's own **document**. 

Here's an example of the original Tome of Beasts document.json file. This is the one that's on the server today.
https://github.com/open5e/open5e-api/blob/staging/data/tome_of_beasts/document.json

Basically, I think the request is:
1. Check the existing spells folder. (we dont' want duplicates). Remove those spells from your list.
  https://github.com/open5e/open5e-api/tree/staging/data/deep_magic
3. For each of the sources of spells, create a new folder in /data/ which includes a document.json describing the document itself. This description can mostly be stolen from the existing data you have.

4. Re-do one PR for each of the new data sources.

Sturlen:
> 1. Remove those spells from your list.

Already done. These are only the spells that I did not find in the existing data.

>  For each of the sources of spells, create a new folder in /data/ which includes a document.json describing the document itself. This description can mostly be stolen from the existing data you have.

I'm happy to split them up, but I'm not sure what documents they came from other than the abbreviated sources I'm fairly certain DM refers to Deep Magic, but what about `DM:RU`? Is this a separate document or part of Deep Magic?


>  Re-do one PR for each of the new data sources.

I don't think this would be helpful, as most of the discussion earlier about the license and sources would apply to all the content equally.

data/kobold_press/document.json
Sturlen: The deep magic document.json also has Open Design LLC as the copyright holder, for the same reason mentioned above.
```
"organization": "Kobold Press",
        "version": "1.0",
        "copyright": "Deep Magic for 5th Edition © 2020 Open Design LLC; Authors: Dan Dillon, Chris Harris, and Jeff Lee."
```

data/kobold_press/document.json
```diff
@@ -0,0 +1,245 @@
...
+            "Deep Magic: Combat Divination © 2019 Open Design LLC; Author: Matt Corley.",
+            "Deep Magic: Dragon Magic © 2017 Open Design; Author: Shawn Merwin.",
+            "Deep Magic: Elemental Magic © 2017 Open Design; Author: Dan Dillon.",
+            "Deep Magic: Elven High Magic © 2016 Open Design; Author: Greg Marks.",
+            "Deep Magic: Hieroglyph Magic © 2018 Open Design LLC; Author: Michael Ohl.",
+            "Deep Magic: Illumination Magic © 2016 Open Design; Author: Greg Marks..",
+            "Deep Magic: Ley Line Magic © 2016 Open Design; Author: Dan Dillon.",
+            "Deep Magic: Mythos Magic © 2018 Open Design LLC; Author: Christopher Lockey.",
+            "Deep Magic: Ring Magic © 2016 Open Design; Author: Dan Dillon.",
+            "Deep Magic: Runes © 2016 Open Design; Author: Chris Harris.",
```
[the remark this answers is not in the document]
eepMoody: To me, at least, that seems like a very reasonable way to split them up. And it also seems like a decision we could revisit if we change our minds and want to get more fine-grained. Shouldn't break anything for items to change their source docs.

I may try and go through and clean up the license sections so they're only listing the actual sources referenced, if you don't mind? As-is they're all very very long, and reference the same list of Kobold Press publications.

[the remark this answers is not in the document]
Sturlen: Feel free. I didn't know if any spells used more outside of the sources they are tagged with, e.g. WotC SRD, so I included everything just in case. I did leave the original source field on each spell in case there needs to be some reshuffling, since most appear in more than one book.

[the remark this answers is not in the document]
eepMoody: Great thinking.

I'm gonna assume for now that the sources they're tagged with are accurate, and as long as we credit at least one source we should be good. Will pull this down over the weekend, cut those down, then get it all merged in! Thank you for all this.

augustjohnson:
**Ok, i'm testing this locally.**
First off: there's a conflict related to some migrations. I think I sorted it out, basically: I removed the help_text on challenge rating and deleted the duplicate 0012 migration. That seemed to work.

**Second issue:**
The spells from DMAG:E don't seem to have a CLASS which is pretty important. They also don't have a school.
The do have a "source" field, which is potentially helpful, but I can't translate the sources to classes/schools.
I am researching this to see if there's a simple mapping to be done.

augustjohnson:
@Sturlen do you have a source document that includes classes and schools (evocation, etc) for these spells that I can manually do the lookups for?

Sturlen:
> do you have a source document that includes classes and schools (evocation, etc) for these spells that I can manually do the lookups for?

I'm afraid not. All the data comes from [this](https://github.com/dimasfh1243/fdrg/blob/main/Foundry%20VTT%20Modules%20for%20DnD%205E/Kobold%20Press%20OGL.md) FoundryVTT plugin, which did not include these fields.

I learned about http://kpogl.wikidot.com/ wiki later, and we might be able the scrape the info from there.

augustjohnson:
Awesome, between those two sources I bet I can get it together. Thanks!

augustjohnson:
OK, @Sturlen I believe it's all looking happy over in:
https://github.com/open5e/open5e-api/tree/kp-spells-cleanup

I think we should close this PR, and I'll make mine from that branch. I had to add a bunch of metadata to a bunch of files, yuck. But it's all set now.

ONE NOTE! Anti-Paladin is the named class on a bunch of spells, and that does NOT map to any of our classes. But we'll get that figured out down the road.

Sturlen:
> I think we should close this PR, and I'll make mine from that branch. I had to add a bunch of metadata to a bunch of files, yuck. But it's all set now.

Sounds good to me.

PR #182, opened 2023-03-26
add separate descriptions for singular and plural endpoints

Dashron:
adds separate endpoint descriptions for singular endpoints i.e. `/monsters/{id}` and plural endpoints i.e. `/monsters`

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -10,7 +10,15 @@ from api.schema_generator import CustomSchema
 
 class ManifestViewSet(viewsets.ReadOnlyModelViewSet):
     """
-    API endpoint for returning a list of of manifests.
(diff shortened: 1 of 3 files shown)


api/views.py
```diff
@@ -10,7 +10,15 @@
 
 class ManifestViewSet(viewsets.ReadOnlyModelViewSet):
     """
-    API endpoint for returning a list of of manifests.
+    list: API endpoint for returning a list of of manifests.
+
+    For each data source file, there is a corresponding manifest containing an
```
Dashron, line 15: heads up, this all falls apart if you use tabs instead of spaces. the retrieve section is viewed as a continuation of the list section

eepMoody (approved):

PR #184, opened 2023-03-26
add oas descriptions for response fields and query params

Dashron:
Adds functional support for response field descriptions and query parameter descriptions.

Next step is to just get all the copy in place. Inline comments on this PR is probably a good way to do that.
- Look at view.py for how you add query param descriptions
- Look at models.py for how you add response field descriptions

I'm not confident this PR will cover all cases, there seem to be some automated fields/params that don't go through the same parsing as the manual ones. As we uncover those I'll find the appropriate ways to incorporate them.

I also noticed that the query parameter for search isn't showing at all, I might need to hack this into schema_generator.py as a one off, like how we delete the `/search/{id}` endpoint.


eepMoody (approved):

augustjohnson:
I'm going to merge and keep my eyes peeled for errors for a bit.

PR #187, opened 2023-03-29
Adding in spells to the tests.

augustjohnson:


api/tests.py
```diff
@@ -71,3 +71,83 @@ def test_options_manifest_data(self):
...
+        self.assertContains(response,'next',count=1)
+        self.assertContains(response,'previous',count=1)
+        self.assertContains(response,'results',count=1)
+
+        # Confirm the values appear as expected.
+        self.assertEqual(response.json()['count'],1)
+        in_mm = json.loads(self.test_spell_json)
+        out_mm = response.json()['results'][0]
+        
+        self.assertEqual(in_mm['name'], out_mm['name'])
```
Gredelston, line 140: Could you put these assertions in a loop? Like:

```py
equal_fields = ('desc', 'higher_level', 'page', 'range', ...)
for field_name in equal_fields:
    self.assertEqual(in_mm[field_name] == out_mm[field_name])
self.assertEqual(in_mm['class'], out_mm['dnd_class'])
```

That will make the test more succinct, and thus easier to read and modify.

api/tests.py
```diff
@@ -71,3 +71,83 @@ def test_options_manifest_data(self):
...
+        response = self.client.get(f'/spells/?format=json')
+        # Confirm the basic elements show up.
+        self.assertContains(response,'count',count=1)
+        self.assertContains(response,'next',count=1)
+        self.assertContains(response,'previous',count=1)
+        self.assertContains(response,'results',count=1)
+
+        # Confirm the values appear as expected.
+        self.assertEqual(response.json()['count'],1)
+        in_mm = json.loads(self.test_spell_json)
```
Gredelston, line 137: What does `mm` mean? I don't think this is semantically meaningful, at least without a clarifying comment.

My best guess is that `mm` could be `monster manual`, but that'd be the wrong source book...

api/tests.py
```diff
@@ -71,3 +71,83 @@ def test_options_manifest_data(self):
...
+        self.assertContains(response,'count',count=1)
+        self.assertContains(response,'next',count=1)
+        self.assertContains(response,'previous',count=1)
+        self.assertContains(response,'results',count=1)
+
+    def test_get_spell_data(self):
+        import json
+        response = self.client.get(f'/spells/?format=json')
+        # Confirm the basic elements show up.
+        self.assertContains(response,'count',count=1)
```
Gredelston, line 130: This is redundant with the assertions in `test_get_spells`.

api/tests.py
```diff
@@ -71,3 +71,83 @@ def test_options_manifest_data(self):
...
+
+    def test_get_spells(self):
+        response = self.client.get(f'/spells/?format=json')
+        # Confirm the basic elements show up.
+        self.assertContains(response,'count',count=1)
+        self.assertContains(response,'next',count=1)
+        self.assertContains(response,'previous',count=1)
+        self.assertContains(response,'results',count=1)
+
+    def test_get_spell_data(self):
```
Gredelston, line 126: nit: Everything here is data, so I'm not sure this test name is meaningfully different than `test_get_spells`.

Gredelston (commented):
minor comments

api/tests.py
augustjohnson: That approach didn't seem to work for me, i was getting a range error, and I realized halfway through that this would result in terminal error messages that would show the VALUE, (i.e. AssertionError 1 !=0) but not have a clue on the NAME of the failed field, so I left it as-is after a bit of cleanup.

I like the idea of succinctness, but not at the cost of troubleshooting context.

Gredelston: This is what the assert methods' `msg` kwarg is for. https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual

> All the assert methods accept a msg argument that, if specified, is used as the error message on failure (see also [longMessage](https://docs.python.org/3/library/unittest.html#unittest.TestCase.longMessage)). Note that the msg keyword argument can be passed to [assertRaises()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises), [assertRaisesRegex()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex), [assertWarns()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarns), [assertWarnsRegex()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarnsRegex) only when they are used as a context manager.

So in this case, it would be like:

```py
self.assertEqual(in_mm[field_name], out_mm[field_name], f'Mismatched field {field_name}')
```

augustjohnson: Beautiful. More commits incoming.

api/tests.py
augustjohnson: Yeah my differentiation here is that get_spells is the list, and testing that the LIST metadata is correct, whereas get_spell_data is testing an individual spell. Probably could be named better, but I'm pushing that one down the road for now.

eepMoody (approved):

PR #189, opened 2023-03-29
Add Descriptions to Tome of Beasts Monsters

### LINKED ISSUE #186: Add Monster descriptions to kobold press monsters
Kobold press includes the monster description in the open content. This ticket is to collect and store those descriptions.

Sturlen:
Fixes #186
Monsters changed:
- ToB1: 349 
- ToB2: 378
- ToB3: 1

Accessible through the `desc` field for the relevant monster, if it had one. No changes to the model were necessary.

All descriptions are from the [Kobold Press FoundryVTT module](https://github.com/dimasfh1243/fdrg/blob/main/Foundry%20VTT%20Modules%20for%20DnD%205E/Kobold%20Press%20OGL.md)

Some escaped unicode characters such were unescaped due to how I processed the json files e.g.  `\u2019` to `’`. Let me know if this is a problem and I'll figure out another workflow.


Gredelston:
This looks great! Thanks for the work here, Sturlen.

AFAIK we don't have an official stance on escaped vs. unescaped characters -- or if we do, it's not documented anywhere. LGTM for now, and maybe we'll standardize later.

Gredelston (approved):

PR #193, opened 2023-04-04
Organization was not imported correctly.

augustjohnson:
Discovered during testing, the ['organization'] value of a document was not imported at all.


Gredelston (approved):
Nice find. I'd love it if we could consolidate all this logic to make issues like this easier to spot, but it's low-priority.

PR #194, opened 2023-04-04
Testing add more tests

augustjohnson:
Added monsters and documents, and moved the imports up to the top.


augustjohnson:
This will fail until #193 is fixed.

augustjohnson:
Added classes and subclasses.

api/tests.py
```diff
@@ -1,5 +1,14 @@
+import json
+
```
Gredelston, line 2: Doesn't matter, but JSYK no newline required here

api/tests.py
```diff
@@ -71,3 +80,274 @@ def test_options_manifest_data(self):
...
+        }
+        """
+        i = Importer(ImportOptions(update=True, append=False, testrun=False))
+        i.import_document(json.loads(self.test_document_json),ImportSpec("test_filename","test_model_class","import_monster"))
+        i.import_monster(json.loads(self.test_monster_json),ImportSpec("test_filename","test_model_class","import_monster"))
+
+    def test_get_monsters(self):
+        response = self.client.get(f'/monsters/?format=json')
+        # Confirm the basic elements show up.
+        self.assertContains(response,'count',count=1)
```
Gredelston, line 163: style nit that doesn't matter: Spaces after commas are not explicitly specified by PEP-8, but they are common in Python, and if we introduce any kind of linting (ex. `pip pep8` or `pylint`) then we'll probably need to add spaces after commas throughout this file.

api/tests.py
```diff
@@ -71,3 +80,274 @@ def test_options_manifest_data(self):
...
+            "armor_desc": "leather armor, shield",
+            "page_no": 315
+        }
+        """
+        i = Importer(ImportOptions(update=True, append=False, testrun=False))
+        i.import_document(json.loads(self.test_document_json),ImportSpec("test_filename","test_model_class","import_monster"))
+        i.import_monster(json.loads(self.test_monster_json),ImportSpec("test_filename","test_model_class","import_monster"))
+
+    def test_get_monsters(self):
+        response = self.client.get(f'/monsters/?format=json')
```
Gredelston, line 161: another style nit that doesn't matter: This doesn't need to be an f-string. I'd recommend just using regular strings unless an f-string is explicitly required.

api/tests.py
```diff
@@ -71,3 +80,274 @@ def test_options_manifest_data(self):
...
+        self.assertContains(response,'previous',count=1)
+        self.assertContains(response,'results',count=1)        
+
+    def test_get_monster_data(self):
+        response = self.client.get(f'/monsters/?format=json')        
+        self.assertEqual(response.json()['count'],1)
+        in_goblin = json.loads(self.test_monster_json)
+        out_goblin = response.json()['results'][0]
+
+        self.assertEqual(in_goblin['name'], out_goblin['name'])
```
Gredelston, line 174: I think it would be much better to consolidate this logic, like:

```py
equal_fields = ('name', 'size', 'type', 'subtype', 'group', 'alignment', ...)
for field_name in equal_fields:
    self.assertEqual(in_goblin[field_name], out_goblin[field_name])

# Some field names unfortunately cannot be identical.
self.assertEqual(in_goblin['speed_json'], out_goblin['speed']

# Some fields are expected to be None
none_fields = ('strength_save', 'dexterity_save', ...)
for field_name in none_fields:
    self.assertNone(out_goblin[field_name])
# then do the same thing for empty strings, too
```

api/tests.py
```diff
@@ -71,3 +80,274 @@ def test_options_manifest_data(self):
...
+        self.assertEqual(in_goblin['speed_json'], out_goblin['speed']) #INPUT IS FORCED TO SPEED_JSON
+        
+        self.assertEqual(in_goblin['strength'], out_goblin['strength'])
+        self.assertEqual(in_goblin['dexterity'], out_goblin['dexterity'])
+        self.assertEqual(in_goblin['constitution'], out_goblin['constitution'])
+        self.assertEqual(in_goblin['intelligence'], out_goblin['intelligence'])
+        self.assertEqual(in_goblin['wisdom'], out_goblin['wisdom'])
+        self.assertEqual(in_goblin['charisma'], out_goblin['charisma'])
+        
+        self.assertEqual(None, out_goblin['strength_save'])
```
Gredelston, line 193: nit: use `self.assertIsNone(out_goblin['strength_save'])`

api/tests.py
```diff
@@ -71,3 +80,274 @@ def test_options_manifest_data(self):
...
+        self.assertEqual(in_goblin['damage_vulnerabilities'], out_goblin['damage_vulnerabilities'])
+        self.assertEqual(in_goblin['damage_resistances'], out_goblin['damage_resistances'])
+        self.assertEqual(in_goblin['condition_immunities'], out_goblin['condition_immunities'])
+        self.assertEqual(in_goblin['senses'], out_goblin['senses'])
+        self.assertEqual(in_goblin['languages'], out_goblin['languages'])
+        
+        self.assertEqual(in_goblin['challenge_rating'], out_goblin['challenge_rating'])
+
+        self.assertEqual(in_goblin['actions'], out_goblin['actions'])
+        self.assertEqual("", out_goblin['reactions']) #Empty string?
```
Gredelston, line 211: Hmm, why are these annotated with question marks? Is this behavior unexpected...?

api/tests.py
```diff
@@ -71,3 +80,274 @@ def test_options_manifest_data(self):
...
+        self.assertEqual("", out_goblin['legendary_desc']) #Empty string?
+        self.assertEqual("", out_goblin['legendary_actions'])#Empty string?
+        self.assertEqual(in_goblin['special_abilities'][0]['name'], out_goblin['special_abilities'][0]['name'])
```
Gredelston, line 214: Could we instead assert that `in_goblin['special_abilities'] == out_goblin['special_abilities']`? Or do we have some weird different data types to deal with?

api/tests.py
```diff
@@ -71,3 +80,274 @@ def test_options_manifest_data(self):
...
+        self.assertEqual(in_document['license'], out_document['license'])
+        self.assertEqual(in_document['desc'], out_document['desc'])
+        self.assertEqual(in_document['author'], out_document['author'])
+        self.assertEqual(in_document['organization'], out_document['organization'])
+        self.assertEqual(in_document['version'], out_document['version'])
+        self.assertEqual(in_document['copyright'], out_document['copyright'])
+
+class CharClassTestCase(APITestCase):
+    def setUp(self):
+        from api.models import Archetype
```
Gredelston, line 265: Move import to top of file

api/tests.py
```diff
@@ -71,3 +80,274 @@ def test_options_manifest_data(self):
...
+
+    def test_get_class_data(self):
+        response = self.client.get(f'/classes/?format=json')
+
+        in_class =  json.loads(self.test_charclass_json, strict=False)
+        out_class = response.json()['results'][0]
+
+        self.assertEqual(in_class['name'], out_class['name'])
+        self.assertEqual(in_class['features']['desc'], out_class['desc']) #The Features key is not duplicated in the api
+        self.assertEqual(in_class['features']['hit-dice'], out_class['hit_dice']) # dashes on entry, underscores out.
```
Gredelston, line 340: If you wind up adding the bulk-assertions pattern that I recommended for Monsters, you could handle this too:

```py
equal_fields = ('hp-at-1st-level', 'prof-armor', ...)
for input_field_name in equal_fields:
    # Input JSON uses hyphens; output JSON uses underscores.
    output_field_name = field_name.replace('-', '_')
    self.assertEqual(in_class[input_field_name], out_class[output_field_name])
```

Gredelston:
Oops, sorry, I'm new to use GitHub for code review (stupid proprietary software at work), so I didn't realize that "add a single comment" would result in this awful spam. Sorry for the spam!

Anyway: God bless you for adding tests. Added some comments, most of which are unimportant and stylistic, but there are a few that I think will make this test a lot more succinct and thus easier to work with.

augustjohnson:
Ok while this isn't perfect, as it turns out neither is our code lol, so this is just definitely better than before. All tests passing at the moment, and a pretty good structure that takes about 90% of Greg's advice above. Once this and the other testing one get merged, I'll start to check off the last couple of object types.

eepMoody (approved):

PR #195, opened 2023-04-05
Schema in docs

augustjohnson:
This adds a schema diagram using mermaid being generated and put into /docs/schema.md. Mermaid diagrams are fully supported by github-flavored markdown, and there's a pretty cool explorable diagram of everything in the database there. I also added a document that more or less says how I created the original diagram in a fairly repeatable format.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -1,0 +1,383 @@
+erDiagram
+
+  api_document {
+    INTEGER id
+    TEXT title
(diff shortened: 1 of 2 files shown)


Gredelston:
Awesome, love to see documentation gains. I'd love it if we could set up automatic regeneration in the future.

Gredelston (approved):

PR #196, opened 2023-04-08
Pipeline in docs

augustjohnson:
Documentation only. Should be mostly accurate. Because of poor committing practice, this should be merged after #195, and duplicated stuff there.


Gredelston (approved):
LGTM! Thank you so much for adding documentation.

PR #201, opened 2023-04-13
a5e - spells

### LINKED ISSUE #161: Add spells from Level Up Advanced 5e
Level up advanced 5e has many spells that both modify existing 5e SRD spells and includes entirely new ones. These are all out under the OGL.

You can find the spells here:

https://a5e.tools/spells

They can likely be scraped or if that proves troublesome, I can reach out to EN World Publishing to see if we can get the database.

Sturlen:
Closes #161

Data is taken from the [Level Up: Advanced 5th Edition (Official) FoundryVTT Nodule](https://github.com/Pjb518/FoundryVTT-Level-Up-Official).

Includes all the normal spell data plus the A5E secondary magic schools, if anyone wants to sort or filter by them in the future.


augustjohnson (changes requested):
These spells do not have a mapping to class. We should try and get that resolved before merge.

augustjohnson:
Ok these are starting to get pretty close.  So bad news is that I've adjusted all of the other spells to include a couple of additional fields (optionally).
1. rolls-attack true (if it rolls attack) for example acid-arrow.
2. saving_throw_ability (a list of abilities that the target must use for saving throws).
3. shape - if the spell effect area has a shape.
> I will take on this task to try and add these fields to these spells based on my parser.

Also, I'm a little concerned about the spell name conflicts. How have re resolved those with the other a5e stuff?

Sturlen:
> * rolls-attack true (if it rolls attack) for example acid-arrow.
> * saving_throw_ability (a list of abilities that the target must use for saving throws).
> * shape - if the spell effect area has a shape.

Added. Also reworked a5e ranges to fit within the new sorting system.



> Also, I'm a little concerned about the spell name conflicts. How have re resolved those with the other a5e stuff?

monsters solve letting you choose their slug, usually by using the document slug as a suffix. spells did not have this function so I added it. Does not affect existing spells which do not use the slug field.

I'd prefer all game content slugs to automatically be suffixed with their document, but this would be a breaking change to the API. Something to consider for V2 though.

augustjohnson:
All great answers. This is on me to review now. Holler at me if I'm taking too long.

augustjohnson:
**OUT OF SCOPE BUT FOLLOW UP TASKS**
- There are no spelllist entries for these items. spelllist is a new concept which ties a spell to a list. Functionally, it's used for things like "wizard spell list" references in something like caster subclasses on martial classes. Also, Herald ~=Paladin, but that's a different name.
- The API has no concept of a5e schools at the moment.  This probably isn't a big deal yet, but worth a follow up eventually.
- Some spells like Writhing Transformation do not conflict with slugs from the wotc-SRD. 
- While I appreciate the data entry on archetypes and such, they are non-functional at the moment. There may be some logical complexities there.  
- We definitely need to test this to confirm that the UI has the proper tag appearing in the lists by class and the list search.

**Things to fix**
- Magic Missile does not have an attack roll.

augustjohnson (approved):

augustjohnson:
To clarify on my "things to fix" it's that magic-missile-a5e['attack_roll'] is set to true, and that's not correct. Magic Missile does not require an attack roll.

PR #202, opened 2023-04-13
Vault of magic additions

mshea:
If someone can run this in a local instance to make sure it works, that'd be great. I don't have a great setup for running a local instance to be sure it's cool.


augustjohnson:
I resolved the conflict.

data/tome_of_beasts_3/vault_of_magic_text_dump.txt
Sturlen, line 1: The source file should probably be contained within the vault of magic folder, not tome of beasts.

mshea: In my local branch it's i a "vault_of_magic" directory underneath the data directory. Is that not how it ended up?

Sturlen: if you look under the Files Changed on Github, you can see that the text file appears to be in the wrong folder:
image
There also seem to be a lot of unrelated files that have been changed too, not sure of the cause.

PR #206, opened 2023-04-15
Vault of magic

mshea:
Added Vault of Magic magic items.


data/a5e_srd/backgrounds.json
Sturlen, line 1: I think you included the a5e backgrounds by accident. they are already in pull request #205, so you can safely delete them from your vault_of_magic branch.

eepMoody (approved):

PR #207, opened 2023-04-17
Spell model upgrade

augustjohnson:
Tons of changes to the spell model, folder structure, data structure. It should be totally backward compatible to API consumers.

**High level list:**
Created an /api/models folder and moved ALL models into there.
Split the Spell model into it's own file.

Created an /api/tests folder and moved all tests in to there.
Created a new /api/tests/test_models.py and added a few spell model tests into there.

**Spell Model adjustments.**
Removed text-based "concentration" field, and added a bool instead.
Added a calculated text-based field to replace.

Removed text-based "ritual" field, and added a bool instead.
Added a calculated text-based field to replace.

Removed a text-based "level" field
Renamed the level_int field.
Added a calculated text-based field for "level" and presented "level_int" as well for backwards compatible.

Removed the text-based component string.
Added 3 booleans related to component requirements.
Added a calculated string to replace the old text-based field.

Added an experimental field called target_range_sort which should allow for tabular spells to be sorted by range.


eepMoody (approved):

api/models/monster.py
```diff
@@ -0,0 +1,120 @@
...
+    charisma_save = models.IntegerField(
+        null=True, help_text='Integer representing the charisma save.')
+    perception = models.IntegerField(
+        null=True, help_text='Integer representing the passive perception score.')
+    skills_json = models.TextField()
+
+    def skills(self):
+        return json.loads(self.skills_json)
+    damage_vulnerabilities = models.TextField(
+        help_text='Comma separated list of damage types the monster is vulnerable to.')
```
Gredelston, line 64: Several of these fields are comma-separated lists. I'd like to explore representing the list more robustly as an actual list of values. You could imagine a scenario where we name a value with a comma in it (ex. "Paralyzed, A5E variant"), and then it's unclear how to parse. We wouldn't have that problem with a bona-fide list value.

I have two suggestions for how to implement this.

1. Use a JSONField to store a list of values. The downside is that the values in that list couldn't be Choices as I described above. We would need to store the string values directly.
2. Create a Model corresponding to each Enum: for example we'd define `class DamageType(models.TextChoices)` and also `class DamageTypeModel(models.Model)`. Then in the Monster model, define columns like `damage_vulnerabilities = models.ManyToManyField(DamageTypeModel)`. I think this is actually a very clean solution: the backend could be very simple, since the enum models never rely on anything else; and from a reader perspective, they would look just like a list of their stringy values. But, it is a little work to refactor.

(Third bad solution, migrate to Postgres, where we have an ArrayField. Lol.)

What do you think? If you think it sounds worthwhile but like a pain-in-the-butt to implement, I could find time for the refactor.

api/models/monster.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def legendary_actions(self):
+        return json.loads(self.legendary_actions_json)
+    spells_json = models.TextField()
+    spell_list = models.ManyToManyField(
+        Spell,
+        related_name='monsters',
+        symmetrical=True,
+        through="monsterSpell")
+    route = models.TextField(default="monsters/")
```
Gredelston, line 104: Huh. Why this field?

api/models/spell.py
```diff
@@ -0,0 +1,229 @@
...
+    # Stored metadata about the spell
+    page = models.TextField(
+        help_text='Page number reference for the document.')
+
+    spell_level = models.IntegerField(
+        validators=[MinValueValidator(0), MaxValueValidator(9)],
+        help_text='Integer representing the level of the spell. Cantrip is 0.')
+
+    dnd_class = models.TextField(
+        help_text='List of classes (comma separated) that can learn this spell.')
```
Gredelston, line 28: This seems like it should be a ManyToManyField to Class.

api/models/spell.py
```diff
@@ -0,0 +1,229 @@
...
+
+    school = models.TextField(
+        validators = [spell_school_validator],
+        help_text='Representation of the school of magic, such as "illusion" or "evocation".')
```
Gredelston, line 32: As I described more thoroughly in `monster.py`, I'm interested in refactoring this a TextChoices field.

api/models/spell.py
```diff
@@ -0,0 +1,229 @@
...
+
+    # Stored data about casting the spell
+    casting_time = models.TextField(
+        help_text='Amount of time it takes to cast the spell, such as "1 bonus action" or "4 hours".')
+
+    range = models.TextField(help_text='Text description of the target range.')
+
+    target_range_sort = models.IntegerField(
+        validators=[MinValueValidator(0)],
+        help_text='Sortable distance ranking to the target.')
```
Gredelston, line 59: I think this help_text is not sufficiently clear for users to understand. Maybe mention what values are assigned for Self, Touch, and others?

api/models/spell.py
```diff
@@ -0,0 +1,229 @@
...
+
+    requires_verbal_components = models.BooleanField(
+        help_text='Casting this spell requires verbal components.')
+    requires_somatic_components = models.BooleanField(
+        help_text='Casting this spell requires somatic components.')
+    requires_material_components = models.BooleanField(
+        help_text='Casting this spell requires material components.')
+
+    material = models.TextField(
```
Gredelston, line 68: I think this is nullable.

api/models/spell.py
```diff
@@ -0,0 +1,229 @@
...
+        help_text='Description of the material required.')
+
+    higher_level = models.TextField(
```
Gredelston, line 71: I think this is nullable.

api/models/spell.py
```diff
@@ -0,0 +1,229 @@
...
+        help_text='What happens if you cast this at a higher level.')
+
+    can_be_cast_as_ritual = models.BooleanField(
+        help_text='Whether or not the spell can be cast as a ritual.')
+
+    # Computed data about casting the spell
+    def v1_components(self):
+        """Presents the components in a standard list like "V, S, M."""
```
Gredelston, line 79: nit/typo: the quotes in "V, S, M" gets confused with the """

api/models/spell.py
```diff
@@ -0,0 +1,229 @@
...
+        help_text='Whether the spell requires concentration')
+
+    def v1_concentration(self):
+        """Field calculated and presented to maintain compatibility with original API."""
+        if self.requires_concentration:
+            return "yes"
+        else:
+            return "no"
+
+    archetype = models.TextField(
```
Gredelston, line 111: As with Class, I think this should be a ManyToManyField.

api/models/spell.py
```diff
@@ -0,0 +1,229 @@
...
+        help_text='Archetype that can learn this spell. If empty, assume all archetypes.')
+    circles = models.TextField(
+        help_text='Druid Archetypes that can learn this spell.')
```
Gredelston, line 114: Why is this separate from Archetypes? Circles are a subset of Archetypes.

api/models/spell.py
```diff
@@ -0,0 +1,229 @@
...
+            if 's' in json['components'].lower():
+                self.requires_somatic_components = True
+            if 'm' in json['components'].lower():
+                self.requires_material_components = True
+
+        if "material" in json:
+            self.material = json["material"]
+
+        # Logic to set boolean based on v1 file import spec (a string).
+        self.can_be_cast_as_ritual = False
```
Gredelston, line 186: I see you're setting a lot of defaults here. Same thing with `requires_verbal_components`, etc. Could we instead include the defaults directly in the fields, like `BooleanField(default=False)`, as you've done for Monsters?

api/models/spell.py
```diff
@@ -0,0 +1,229 @@
...
+        if "ritual" in json:
```
Gredelston, line 187: nit: L187-193 can be compressed:

```py
self.can_be_cast_as_ritual = json.get("ritual", None) == "yes"
```

api/models/spell.py
```diff
@@ -0,0 +1,229 @@
...
+            # Default is false
+            if str(json["ritual"]).lower() == 'yes':
+                self.can_be_cast_as_ritual = True
+
+            if str(json["ritual"]).lower() == 'no':
+                pass  # Already set as False through default.
+
+        if "duration" in json:
```
Gredelston, line 195: In fact, there's lots of fields that can be compressed using `.get()`!

`self.duration = json.get("duration", None)` (or whatever default value we want)

We also have an option to dynamically fetch the default value:

```py
self.duration = json.get("duration", self._meta.get_field("duration").default)
```

That's a little ugly, but we can always make an abstraction!

```py
def set_if_present(json: Dict[str, Any], field: str):
    default = self._meta.get_field(field)
    setattr(self, field, json.get(field, default)))
```

...Eh, maybe that's not cleaner. I think the first snippet I wrote is cleanest. Anyway, just wanted to expose you to the option of `.get()` in case you hadn't considered it. Not required.

api/models/spell.py
```diff
@@ -0,0 +1,229 @@
...
+            self.duration = json["duration"]
+
+        # Logic to set a boolean based on v1 file import spec (a string).
+        self.requires_concentration = True
```
Gredelston, line 199: Typically False is the default value for bool fields. I recommend sticking with that convention.

api/models/spell.py
```diff
@@ -0,0 +1,229 @@
...
+
+    dnd_class = models.TextField(
+        help_text='List of classes (comma separated) that can learn this spell.')
+
+    school = models.TextField(
+        validators = [spell_school_validator],
+        help_text='Representation of the school of magic, such as "illusion" or "evocation".')
+
+    # Computed metadata about the spell
+    def v1_level(self):
```
Gredelston, line 35: I'm not sure whether this is helpful or not, but we could consider decorating these with `@property` so that we can access them like regular fields.

In other words, instead of `my_spell.v1_level()`, we could do `my_spell.v1_level`. All you'd need to do is put `@property` immediately above the `def` line.

Not a big deal, just offering.

api/views.py
```diff
@@ -94,23 +94,28 @@ class DocumentViewSet(viewsets.ReadOnlyModelViewSet):
         )
 
 class SpellFilter(django_filters.FilterSet):
+    # TODO the filter for level_int was lost in the modelfield rename, replace this.
```
Gredelston, line 97: I see a `level_int` filter. Is this TODO resolved? If not, what needs to be done?

data/tome_of_beasts_3/monsters.json
```diff
@@ -12637,7 +12637,7 @@
   "page_no": 168
  },
  {
-  "name": "Angel, Archangel Iilari\u2019Jil",
+  "name": "Angel, Archangel IilariJil",
```
Gredelston, line 12640: Here and on L23100, it looks like you've accidentally removed the ’ character. (\u2019 is the escape code for right single quotation.) If your intention was to remove the escape character, maybe we could use an apostrophe instead? Iilari'Jil would be more proper.

Gredelston (commented):
Nice work! Glad I finally got through the review. I think you've laid some really solid scaffolding for future improvements.

My biggest comment here is advocating for the use of Choices and ManyToManyFields. I know we went back-and-forth on that in 1:1 chat on Discord. If it seems like a pain in the butt to implement, let me know and I can hack on it.

api/models/models.py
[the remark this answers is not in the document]
augustjohnson: Except for the fact that I tore out Monster and MonsterSpell (but didn't edit them), yes.

api/models/monster.py
```diff
@@ -0,0 +1,120 @@
...
+        help_text='Short description of the creature alignment, such as "lawful good"')
+    armor_class = models.IntegerField(
+        default=12, help_text='Integer representing the armor class.')
+    armor_desc = models.TextField(
+        null=True, help_text='Description of the armor or armor type.')
+    hit_points = models.IntegerField(
+        null=True, help_text='Integer of the hit points.')
+    hit_dice = models.TextField(
+        help_text='Dice string representing a way to calculate hit points.')
+    speed_json = models.TextField()
```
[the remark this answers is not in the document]
augustjohnson: This is generally out of scope, I didn't touch Monster, but I'll keep this in mind.

api/models/monster.py
```diff
@@ -0,0 +1,120 @@
...
+        help_text='The type of the monster, such as "aberration"')
+    subtype = models.TextField(
+        help_text='If applicable, the subtype of the monster, such as "shapechanger"')
+    group = models.TextField(
+        null=True,
+        help_text='Used to group similar creatures at different stages. "Green Dragon"')
+    alignment = models.TextField(
+        help_text='Short description of the creature alignment, such as "lawful good"')
+    armor_class = models.IntegerField(
+        default=12, help_text='Integer representing the armor class.')
```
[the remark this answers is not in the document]
augustjohnson: Eh yes, monster needs work. Out of scope of this PR though.

api/models/spell.py
augustjohnson: Yes, I generally agree. I think there's a bit of follow up work to be done on spell that includes at a minimum bumping this out into a list, but there's implications on loading that I'm not sure I can sort out easily. Your concerns is understood, and I agree that long-term it's the best decision. Out of scope of this larger PR though.

api/models/spell.py
augustjohnson: Not from what I can tell:
https://docs.djangoproject.com/en/4.2/ref/models/fields/#null

api/models/spell.py
augustjohnson: I don't believe so.
https://docs.djangoproject.com/en/4.2/ref/models/fields/#null
Some of this gets weird because we ignore the CUD of CRUD, so while it may be nullable, it's impossible to update to that value.

api/models/spell.py
augustjohnson: Nice catch, fixed.

api/models/spell.py
augustjohnson: Yeah I think there's a data model discussion to be had here. What's the difference between an archetype and a circle? If I list both Cleric and Life Domain, what does that actually mean?
For now I'm assuming archetype and circle are literally unused, and it's a problem to be solved at a later date.

api/models/spell.py
augustjohnson: Yeah that needs resolution. Not in scope for this PR though.

data/tome_of_beasts_3/monsters.json
augustjohnson: Yep, thank you.

api/models/spell.py
augustjohnson: Yes, I've set all of the booleans to default as false.

api/models/spell.py
augustjohnson: I have adjusted all of the imports to be simpler, but I didnt' quite one-liner them. Some of them I'm parsing text, so two lines feels more natural to me, but It's cleaner than it was based on your input.

api/models/spell.py
augustjohnson: Good advice, I'm ignoring at the moment, but filing it away for future improvements.

api/models/spell.py
augustjohnson: Changed.

api/views.py
augustjohnson: It works automatically! Django knows that level_int is an alias of spell_level.

api/models/spell.py
augustjohnson: I see these fields as vestigial mostly, and not worth encouraging access to, especially not in code.

api/models/spell.py
augustjohnson: Added help text with values.

api/models/spell.py
augustjohnson: Yes, this will be incoming relatively soon.

augustjohnson:
I incorporated MOST of the feedback on spells, and ignored the feedback on monsters. I only broke monsters out for a bugfix reason while I was splitting spells out, I didn't touch it other than that.

More changes ARE needed, but this gives a reasonable foundation.

Gredelston (approved):

PR #210, opened 2023-04-22
Kobold Press Spells Merge Take 2

augustjohnson:


augustjohnson:
This included all spells in Sturlen's original branch, and I auto-manually mapped them to classes and schools based on data found in these two sources.
http://kpogl.wikidot.com/
http://kpogl.wdfiles.com/local--files/home:home/koboldpressogl.zip

I did a name-based lookup. It did result in a couple of spells that did NOT have classes, but they were all in the Warlock folder, so I assumed they were solo warlock spells.

api/models.py
```diff
@@ -127,7 +127,7 @@ def skills(self):
     condition_immunities = models.TextField()
     senses = models.TextField()
     languages = models.TextField()
-    challenge_rating = models.TextField(help_text='Monster challenge rating.')
+    challenge_rating = models.TextField()
```
eepMoody, line 130: It's taking this help text out intentional?

data/warlock/document.json
```diff
@@ -0,0 +1,245 @@
...
+            "12. Inability to Comply: If it is impossible for You to comply with any of the terms of this License with respect to some or all of the Open Game Content due to statute, judicial order, or governmental regulation then You may not Use any Open Game Material so affected.",
+            "13. Termination: This License will terminate automatically if You fail to comply with all terms herein and fail to cure such breach within 30 days of becoming aware of the breach. All sublicenses shall survive the termination of this License.",
+            "14. Reformation: If any provision of this License is held to be unenforceable, such provision shall be reformed only to the extent necessary to make it enforceable.",
+            "15. COPYRIGHT NOTICE",
+            "Open Game License v 1.0a Copyright 2000, Wizards of the Coast, Inc.",
+            "12 Perilous Towers © 2018 Open Design LLC; Authors: Jeff Lee.",
+            "A Drinking Problem ©2020 Open Design LLC. Author Jonathan Miley.",
+            "A Leeward Shore Author: Mike Welham. © 2018 Open Design LLC.",
```
eepMoody, line 41: I think we should slim down these source docs to only the ones that are relevant, eg. The various Warlock publications for this set?

Happy to hop on this myself this morning, if you'd like.

eepMoody (approved):
A couple of thoughts, but nothing blocking!

api/models.py
augustjohnson: Yes, this caused a migration to be created, then a migration conflict. That conflict was resolved elsewhere, but the model didn't get dleted.  That line is totally rewritten elsewhere anyway, so basically, it's a very temporary reversion of something that shouldn't have been in this PR anyway.

data/warlock/document.json
augustjohnson: I agree, but that can be done transparently after-the-fact, so it wasn't a big priority for me.

PR #222, opened 2023-05-12
Spell class list endpoint and data.

augustjohnson:
To test this:
1. I added a bunch of tests, both on the model side, as well as the API side. Certainly not all cases, but the happy path is tested.
2. I did manage.py quickload successfully.
3. I did manage.py test
4. I reviewed the o5e spelllist.json file compared to the spells.json file for accuracy. It was good to go.


eepMoody (approved):

eepMoody:
Very cool and I like this solution! Makes them easy to adjust, and the generation script will be perfect for future content.

PR #235, opened 2023-05-16
Removing a single typo character in the wish spell

augustjohnson:
Literally a 1 character change, from Yyou to You.


eepMoody (approved):
I dunno, this one has such massive scope that I can't be confident I've thoroughly reviewed it. But I guess we just have to hope. 😆

PR #239, opened 2023-05-17
Task/install pytest

guizesilva:
The goal of this PR is to install pytest so it can be used as the test runner.

Pytest offers some benefits over the default Django's test runner, but is totally backwards compatible with it.

Some of the features offered by pytest:

- [Manage test dependencies with pytest fixtures.](https://pytest.org/en/latest/how-to/fixtures.html)
- Less boilerplate tests: no need to import unittest, create a subclass with methods. Write tests as regular functions.
- Database re-use: no need to re-create the test database for every test run.
- Run tests in multiple processes for increased speed (with the pytest-xdist plugin).
- Make use of other [pytest plugins](https://pytest.org/en/latest/how-to/plugins.html).
- Works with both worlds: Existing unittest-style TestCase's still work without any modifications.

This might be a good start to make writing tests easier and increase the test coverage of the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ The Django API uses Django REST Framework for its browsability and ease of use w
     are on the [pipenv website](https://pipenv.readthedocs.io/en/latest/).
 
 1.  Once pipenv is installed, you can install all of the project dependencies
-    defined in the Pipfile via `pipenv install`.
+    defined in the Pipfile via `pipenv install --dev`.
 
 ## Quick Setup
 
--- a/README.md
+++ b/README.md
@@ -52,6 +52,13 @@ At that point, you will be able to run the django server normally (within the pi
 
 And your server should be available at http://localhost:8000.
 
+## Tests
+
+To run the test suite:
+```
+pipenv run pytest
+```
+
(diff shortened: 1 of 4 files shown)


augustjohnson:
Things I like about this:
1. This introduces the concept of dev-packages which is pretty important.
2. It's summary is better than unittest.
3. According to this google trend, pytest seems to be the way to go: https://trends.google.com/trends/explore?date=today%205-y&geo=US&q=python%20pytest,python%20unittest,python%20nose&hl=en
4. It runs all of the existing tests.

Things I don't like about the PR as-is:
1. The instructions on running tests are a bit weak and I had to look through them a couple of times.  Think you could update the README with a very clear section from checkout to running tests?

README.md
augustjohnson, line 1: Please add better instructions for running the tests.

guizesilva: hi @augustjohnson , thanks for the review.

I added additional instructions for installing the requirements before running the tests.
Is there anything else that you felt the instructions were lacking when you tried to run the tests?

augustjohnson (approved):

PR #242, opened 2023-05-18
Disable Django Migrations

Sturlen:
Reduce build times and merge conflicts by removing all existing migrations and disabling the feature in the django config. 

Migration files are important when working with a persistent database, but serves no purpose when we start from a fresh database every build. 

Changes have been built and tested locally, as well as in production mode live at: https://open5eapi.spetland.no .

Build functions the same, but any existing local migration files should be deleted.


eepMoody (approved):
This looks good to me, and can confirm it works fine locally. @augustjohnson any 🚩 s jumping out to you? Otherwise I think we can merge this one in.

augustjohnson:
This is scary.
Ok things I like:
1. Deletion of a bunch of auto-generated cruft that was functionally not needed.
2. "Migrations" run faster in quicksetup now.

**Things I don't like:**
If someone does model work, adding a field for example, the LOE to get the path to get it working is now not as well worn as manage.py makemigrations / migrate, which is the answer the google will have. Those fail with unhelpful errors.

In fact, I haven't found a simple way to get it to work without a delete database and run quicksetup again.

I'm holding up a stop sign to this at the moment just because the status quo is working, and this doesn't provide a significant benefit.

augustjohnson (changes requested):
See comment.

Sturlen:
> If someone does model work, adding a field for example, the LOE to get it working is now not as well worn as manage.py makemigrations / migrate, which is the answer the google will have. Those fail with unhelpful errors.
> 
> In fact, I haven't found a simple way to get it to work without a delete database and run quicksetup again.

In this project, the DB is essentially a build artifact and not something that should be persisted between builds. Therefore, rebuilding the DB should be the norm when you're modifying fields. This also makes sure there is no leftover data from a previous build.

As for the googling, I think this could be improved by removing the old DB as part of quicksetup and eliminating the need to run the individual commands.

In the end I still think removing migrations will help in the long run by making PR's mergeable in any order without human intervention.

augustjohnson:
The **positive** of this change is that upon committing an edit of a model, there would no longer be a possibility of a merge, or out-of-order migration.
The **negative** of this change is that each edit of a model now requires a ~200 second rebuild to test rather than being a ~2 second migration.

In my mind the negative far outweighs the positive of this, but I realize that may the the only one impacted by the negative. Is the positive that much better for the group?

augustjohnson:
Just as a note, the rebuild of the indexes is the piece that takes like 195 seconds out of those 200. Which isn't always required. We could maybe disable this like and then have a noindex param or something that results in the QUICK automigrate. I'm still not totally convinced this is better though, even if we had noindex as an option.

eepMoody:
I've added a --noindex param to quicksetup that I believe reduces the rebuild to the target time we were looking for? Takes ~10 seconds on my computer at this point.

Do we want to close this? Or is it something we still want to keep thinking on?

Sturlen:
Building with `--noindex` now solves most of the negatives. This should probably be combined with an update to the readme for how to work with the project.

eepMoody:
So should we close this down and make those updates to the readme separately? It feels like we've decided this isn't a route we want to go down.

augustjohnson:
Closing.  We can reopen if necessary.

PR #245, opened 2023-05-19
exposed monster enviroments to the api

tbasten:
Added the new monster environment tags to the api


augustjohnson (changes requested):
The environments field needs to be returned as a json list. We've messed this up a few times in the past, I'm determined not to do it again.

See https://github.com/open5e/open5e-api/blob/staging/api/models/monster.py#L81 for inspiration on how to take a string as input and return it as a list.

augustjohnson:
I merged this in, I'm going to see if I can get it across the finish line in json format.

augustjohnson:
This is currently available on the production API.

PR #248, opened 2023-05-20
Introduce versioning to DRF API

dprothero:
This PR introduces the concept of versioning to the API and sets the current state of the API as `v1`, making that the default version if no version is passed in the URL.

[See this reference](https://www.django-rest-framework.org/api-guide/versioning/)

One thing to keep in mind when it comes time to add a new version, you can reuse Views in your urls if they should just be the same as the prior version.


augustjohnson:
This works as expected I think. There could be some breakage maybe? But it seems to be designed to avoid that. I approve moving it on.

augustjohnson (approved):

PR #259, opened 2023-05-28
Create races.json

JonathanELawrence:
Added races & subraces + Errata 02/01/2023

### DIFF
--- a/api/management/commands/quicksetup.py
+++ b/api/management/commands/quicksetup.py
@@ -1,4 +1,5 @@
 """Helper command to fully set up the API."""
+import argparse
 
 from django.core.management import call_command
 from django.core.management.base import BaseCommand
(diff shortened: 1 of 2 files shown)


eepMoody:
So on running this through our importer, and testing on the local server I got an interesting error:

`JSONDecodeError Expecting value: line 1 column 1 (char 0)`

Did you get this when you ran it locally? I have a suspicion based on some research that it's coming from an empty string somewhere, but can't nail it down. @augustjohnson do you have any thoughts on the best way to diagnose?

JonathanELawrence:
I think it's "traits": "***Hunger for Flesh.*** it appears grey rather than colored. Not sure exactly what.

eepMoody:
Been doing some digging this morning, and there are quite a few cases of issues.

I fixed some of the easier ones to resolve, but the big problem is the existing races getting new subraces. We don't really have a mechanism for doing that, so they're overwriting races like "Gnome" and creating invalid races (no ASIs, no speed, etc.)

I think maybe we want to file that as a new issue, and come up with a format for appending sub-races to existing ones.

eepMoody:
I've made a PR against your branch for this, so you can see what's being taken out. It's a few pretty big chunks, and I had to add a few values for things like speed that were redundant (for some subraces) or potentially overridden by a trait (shade). Haven't had a race with truly variable speed before, so we probably want to create some issues to handle these cases we've uncovered:

- Ability to add a subrace to a race from a different source
- Ability to accept a subrace with no explicit speed (base it on the base race in this case)
- Ability to accept a subrace with no ASI (can be blank in this case)
- Ability to set variable ASI (currently using "Any" but cases like "Any but Constitution" exist and not sure how to represent that)

augustjohnson:
Broadly these issues are intended to be addressed by my concept of Feature here:
https://github.com/open5e/open5e-api/blob/api_v2_design_docs/docs/v2/models.md

But that doesn't solve a short-term problem.

eepMoody (approved):

PR #261, opened 2023-05-28
Create feats.json

JonathanELawrence:
Added feats from Tome of Heroes.


eepMoody (approved):
Built and tested locally, all working great!

PR #262, opened 2023-05-29
Create magicitems.json

JonathanELawrence:
Added magic items from Tome of Heroes by Kobold Press


eepMoody (approved):
built and tested locally, looks great!

PR #263, opened 2023-05-30
task: Add basic tests and fixtures for the spells view.

guizesilva:
This PR is implementing basic tests for the Spells view.

A few pytest features are also being introduced:

- We're making use of [fixtures](https://docs.pytest.org/en/6.2.x/fixture.html):
    - A "root" fixture is being created for the Document model;
    - A module level fixture is created for the Spell model;
    - The document fixture is used on defining the spell fixture, and then the spell fixture is used inside the test function;
    - We use the client fixture for making requests (made available by the [`pytest-django`](https://pytest-django.readthedocs.io/en/latest/helpers.html#client-django-test-client);
   
- Simple [`assert`](https://docs.pytest.org/en/6.2.x/assert.html#asserting-with-the-assert-statement) statements are used for testing the expected conditions.
- A [`mark`](https://docs.pytest.org/en/6.2.x/assert.html#asserting-with-the-assert-statement) is used to allow db access inside tests.

### DIFF
--- a/.github/workflows/pr_validation.yml
+++ b/.github/workflows/pr_validation.yml
@@ -18,11 +18,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Install Dependencies
-        run: pipenv install
+        run: pipenv install --dev
       - name: Run migrations
         run: pipenv run python manage.py migrate
       - name: Run Tests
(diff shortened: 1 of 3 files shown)


eepMoody (approved):
nice 💯

PR #265, opened 2023-06-03
Create spells.json

JonathanELawrence:
Added all Tome of Heroes spells + Errata 02/01/2023


eepMoody:
Did a manual merge on this one because it got a little messy! Was missing a 1-line change from an earlier import of the same spells.

PR #270, opened 2023-06-08
Backporting changes into my v2 branch.

augustjohnson:


augustjohnson:
Deleting, this is not helpful.

Issue #250, opened 2023-05-21
Add telemetry or tracking to API


eepMoody:
Currently, there is very little information on how our API is used. Our website has excellent analytics through Google Analytics, but the API consumers likely have very different behavior.

Some initial investigation suggests that [DRF Tracking](https://drf-tracking.readthedocs.io/en/latest/) is the standard, and is a safe place to start.

augustjohnson:
Doesn't the gunicorn log have all of this information already? I suppose there's some parsing to do, but if we're implementing at the API layer, the log is already there, just ignored.

I think one thought is: 
Write the gunicorn log to digitalocean's storage.
Use Papertrail / Logtail (add-on) on the free tier to review logs.
https://cloud.digitalocean.com/add-ons?i=a3593f

The issue with this approach is that it does not get a realistic picture of usage becuase of cloudflare's caching. Gunicorn doesn't even see the majority of requests (hopefully).

We have to get this info from cloudflare. We should poke around on there to make sure we understand what it collects and what it can.

eepMoody:
There is now telemetry being recorded via NewRelic, which provides more than enough free scale to serve our needs!

PR #272, opened 2023-06-19
Add telemetry and hide v2 apps

eepMoody:
Trying out New Relic telemetry, which should cover our monthly usage on their free plan, with enough logins to allow several devs to monitor usage and post insights for general consumption.


eepMoody:
closed #250

PR #275, opened 2023-06-20
add serializer methods to avoid errors on json-based fields

eepMoody:
Fixes errors like

` django.core.exceptions.FieldError: Cannot resolve keyword 'reactions' into field. Choices are: actions_json...`


eepMoody:
self-approving this one so we can get a cut to main ready. Have tested multiple times on local and remote, with every possible combo of filters and sorting I could come up with.

PR #287, opened 2023-06-28
hocus pocus markus downus

bradyslot:
alakazam!


data/deep_magic_extended/spells.json
```diff
@@ -812,7 +812,7 @@
     },
     {
         "name": "Snowblind Stare",
-        "desc": "Your eyes burn with a bright, cold light that inflicts snow blindness on a creature you target within 30 feet of you. If the target fails a Constitution saving throw, it suffers the first stage of snow blindness (see [[[Conditions]]]), or the second stage of snow blindness if it already has the first stage. The target recovers as described in [[[Conditions]]].",
+        "desc": "Your eyes burn with a bright, cold light that inflicts snow blindness on a creature you target within 30 feet of you. If the target fails a Constitution saving throw, it suffers the first stage of snow blindness (see [Conditions](https://open5e.com/sections/conditions)), or the second stage of snow blindness if it already has the first stage. The target recovers as described in [Conditions](https://open5e.com/sections/conditions).",
```
eepMoody, line 815: Hey @bradyslot, we may need to roll back some of these changes that create hard links to the open5e site. 

One of our bits of design philosophy has been to absolute links to our URLs, since the consumer might want to link to their own version of the resources (or do something else with the link).

Finding a good mechanism for cross-linking has been a back-burner project for a while though, if you have ideas?

eepMoody (approved):
Other than the explicit links to open5e.com, this all looks awesome to mee!

data/deep_magic_extended/spells.json
bradyslot: That's a sound philosophy.
I don't have a solution for that as yet.
Wait to merge until I think of something 🤔

bradyslot: I think for now having
`[Spell Name]({{ base_url }}/spells/spell-name)`
solves browsability for consumers willing to implement jinja2 parsing.
It's certainly better off in this form than what existed before, where browsing wasn't really possible.
I don't know how to solve forward compatibility with unique slugs in v2.
Any ideas?

eepMoody (approved):
Approved after some discussion. Spun off #291 and #292  to deal with downstream impacts.

Thank you for this @bradyslot, sorry for the noise. Just have to be deliberate about setting a new standard!

PR #302, opened 2023-07-19
Api v2 items

augustjohnson:
This results in v2 being turned on, and a few endpoints becoming available.  Based on my understanding of the usage patterns there won't be any failed requests.

**I added:**
New endpoints:
* /v2/items most notably for a view of all types of items.
* /v2/weapons for a definition of a weapon.
* /v2/armor - similarly for armor
* /v2/itemsets - a list that shows a collection of items.
* /v2/documents, /v2/licenses, /v2/publishers for a new structure related to v2 content.
* /v2/rulesets which a document can be published under. 

I re-organized and imported a ton of existing data.
All magic items from the wotc-srd
Almost all magic items from kobold-press vault of magic. - exceptions can be fit into bugs.
Various non-magical items from the SRD such as potions, trade goods, ammunition, and a few other things that were poorly represented now have an entry in the items table.

**I tested:**
> adding a few tests to ensure that the /v2 endpoint does show up.
> hitting /v1/weapons (and getting the v1 endpoint).
> hitting /weapons (and getting the v1 endpoint).
> hitting /v2/weapons (and getting the v2 endpoint).
Same for /armor.

This is ready for BETA usage at this time I think, but breaking changes certainly still could occur.


augustjohnson:
After a review with E, this here are the 3 changes requested.
1. Change itemsets singletons to return all item values.
2. Expose the Key field
3. Figure out a easy deployable way for v2 to be enabled on staging, but not production.

augustjohnson:
To resolve 1. above, I have added a parameter on all gamecontent items that takes "depth=1" and would then follow the serializer for the referenced object below. - The object below though, does NOT include calculated fields, so I'm working on that.
To resolve 2, I have added the key parameter on all objects.
I haven't yet resolved 3.

augustjohnson:
1 2 and 3 are resolved above, I've added ?depth=1 but it's not as useful as it seems. I think it's all clean and ready to go.

eepMoody (approved):

PR #304, opened 2023-07-26
a5e magic items

Sturlen:
Parsed from https://a5e.tools/magic-items

Contents still need to be checked for errors and verified, which is why I'm putting this a draft for now

### DIFF
--- a/data/a5e_srd/magicitems.json
+++ b/data/a5e_srd/magicitems.json
@@ -1,0 +1,5462 @@
+[
+    {
+        "name": "Iron Flask",
+        "slug": "iron-flask-a5e",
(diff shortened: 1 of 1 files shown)


Sturlen:
Items should now be up to current Open5e spec. 
The fields "requirements", "components" and "cost" are not currently exposed in the API, but I intend to make a separate PR for these.
"source" and "source_href" have been left in to make it easier to point to a specific source book at a later time.

This content update includes 545 items. I've tried my best to skim through and check for any errors, but with so many entries I have probably missed something.

eepMoody:
Hey @Sturlen taking a look at this this week, but it's a big PR so it's just taking a while. Haven't forgotten about it!

eepMoody (approved):

PR #309, opened 2023-08-03
Bugfix for array properties becoming an empty string when empty 

daydin:
there might be more properties that need the same fix in the same file, needs a second look by a mod.
Namely, L436-L451: shouldn't every if be matched by an else that dumps a None to the JSON? @eepMoody 
Discussion here: https://discord.com/channels/470638316103008268/1120433155841728613/1120433155841728613

### DIFF
--- a/api/management/commands/importer.py
+++ b/api/management/commands/importer.py
@@ -461,7 +461,7 @@ class Importer:
                 monster_json["actions"][idx] = z
             i.actions_json = json.dumps(monster_json["actions"])
         else:
-            i.actions_json = json.dumps("")
+            i.actions_json = json.dumps(None)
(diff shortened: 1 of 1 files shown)


augustjohnson:
This is a bit of a change, i'm hoping there's not downstream problems. YOLO I suppose.

augustjohnson (approved):

daydin:
> This is a bit of a change, i'm hoping there's not downstream problems. YOLO I suppose.

Can you allow me to merge? Right now I have no write access it seems.

augustjohnson:
Since merging results in an automated deploy, that is generally done in batches, and done by the maintainers (myself and @eepMoody).  We'll queue this up next release.

PR #311, opened 2023-08-04
open5e original feat "Magical Novice"

nstefanski:
Replacement for a standard feat not in the 5.1 SRD.

I wanted to help with original content, starting with something small. Let me know what you think, if this is distinct enough or if it needs further edits.


augustjohnson:
So I think this may be a bit of a conflict with the a5e SRD.
Here's the equivalent feat that's already from LevelUp 5e, which we're trying to pull in all of their stuff.

https://open5e.com/feats/mystical-talent

At this point, I'd vote NOT to include this additional data mostly because it's adding a third name and source for a feat that most people will know as Magic Initiate.

nstefanski:
I missed that one from A5e, I admit I'm not as familiar with their content as Kobold Press. I'll double check the A5e list to avoid duplicate content.

PR #312, opened 2023-08-05
166 consider adding detailed code version information in an api response

augustjohnson:


augustjohnson:
This is not absolutely confirmed to work, since it's a script intended to only be run on public deployments.

eepMoody (approved):

eepMoody:
Approved, but looks like something is causing a build fail on this one.

Issue #301, opened 2023-07-17
When searching for feats it always returns every feat


JeremyMees:
I make this api call `https://api.open5e.com/feats/?limit=20&search=slasher&page=1`
And it returns all the feats instead of filtering them on name. I use the same logic for every type and i only have problems with the feats.

augustjohnson:
I can't find a simple reason for this:
Research done:
Tried to recreate the issue locally: Failure
Tried to recreate this issue on a few different models: Fail
Read up on:
https://www.django-rest-framework.org/api-guide/filtering/#searchfilter
- Search should NOT be behaving this way from what I can tell.
Confirmed search_fields are set correctly.
https://github.com/open5e/open5e-api/blob/staging/api/views.py#L261

So from what I can tell this is a caching issue, and my prediction is that cloudflare is set up in a funky way with the ?search query param... but I'm not certain. I don't have access to cloudflare, so I'll have to ask @eepMoody to take a look.

augustjohnson:
Checking back on this:
The original error scenario of https://api.open5e.com/feats/?limit=20&search=slasher&page=1 no longer seems to return ANY results. In fact, I can't seem to find a feat named slasher at all in our history? I'm guessing that this can reasonably closed as couldn-not-recreate :/  Sorry!

PR #313, opened 2023-08-05
Add search fields to endpoints that lack them

greshdev:
Ensured that all `ViewSets` (aside from Manifest, Search, User and Group) had search fields for name and description (`DocumentViewSet` uses `title` instead of `name`). This involved adding `search_fields` for `name` and `desc` to all `ViewSet`s that lacked search fields and adding a search field for `desc` to all `ViewSet`s that only had `name`. 
This should fix the bug identified in issue #301.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -605,4 +605,4 @@ class ArmorViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = models.Armor.objects.all()
     serializer_class = serializers.ArmorSerializer
     filterset_class = ArmorFilter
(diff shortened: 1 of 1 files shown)


augustjohnson (approved):

augustjohnson:
Hey @greshdev I just merged in #314 and I think that's causing the conflicts on this. Can you resolve them and add another commit?

Thanks!

greshdev:
Thanks! I think that should fix it, let me know if there's still any issues.

PR #317, opened 2023-08-13
Implement initial Creature model, including Abilities

wardsky:


augustjohnson:
This is definitely on the right track, while creatures still need work to be functional, I think what you've laid out here is a great skeleton. Approved.

augustjohnson (approved):

eepMoody:
Sorry it took a bit to get to this, had some life stuff come up (plus baldur's gate 😆 ) but this is amazing stuff. Really cool work, super excited to see where it's going and we've been talking about how incredible this is going to be once it's live.

PR #319, opened 2023-08-15
Api v2 feat race

augustjohnson:
This adds feats and races to /v2/, and includes all data currently available on /v1/.
It actually includes more feats, because our /v1/ list of a5e feats was incomplete compared to what they released. This list should be complete based on a5e's currently released feats.


augustjohnson:
Rescinding this PR for a bit, I've got some more work to do.

PR #320, opened 2023-08-18
Fix CR values for TOB3

Thomvis:
TOB3 is the only document with numerical CR values (e.g. `0.25`). All other documents use a string representation of a fraction (e.g. `"1/4"`).


augustjohnson:
Good find.

augustjohnson (approved):

PR #334, opened 2023-10-11
Add api and data version hashes

Sturlen:
As discussed on Discord. Walks and hashes the contents of the API and data folders on startup. I believe this to be more accurate than relying on git commit hashes. This is intended to replace /manifest as the way to notify a client about changes. 
image


augustjohnson (commented):
Ok this is way better than the janky non-functional thing that I had, but it doesn't tie back to a git commit.  Any chance we could add a git commit hash, and maybe a build ID into this endpoint?

augustjohnson (commented):
Ok you have some issues here. 

First you're calling your variable SHAhash even though it's not using SHA.

Second, md5 shouldn't be used really ever. I'd suggest replacing all usage with SHA256, or maybe using the hashlib.file_digest feature, which may speed some things up, but I would guess this is already snappy.

Third, I believe your with open statement could be simplified and clarified to:
digest=hashlib.sha256()
loop through files here:
    with open(filepath, 'rb') as f1:
        digest = hashlib.file_digest(f1,digest,"sha256")

return digest

Sturlen:
> Any chance we could add a git commit hash, and maybe a build ID into this endpoint?

Git commit is in now. where would you get a build ID from, and what purpose would it serve?

augustjohnson (commented):
I don't think subprocess git will work in production, the python alpine image does not include (and should not include) git.
Let's set aside the git commit info. The rest is great.

augustjohnson:
I have included this in the branch release/1.4.0

augustjohnson:
Available here.
https://api-beta.open5e.com/version/

Issue #335, opened 2023-10-11
Monstrous Menagerie bonus/reactions not showing up correctly.


augustjohnson:
Per kregk in a discord issue:
 I'm finding a lot of monsters from monstrous menagerie seem to be missing bonus actions and skill proficiencies (or bonus action comes up as reaction instead). (Adult river dragon, azer, boggard, bone devil, cambion, etc.) I've been creating/fixing statblocks for my own use using https://dnd-statblock.crapmybrainsays.com/  I don't anything about coding, so I don't know how I can fix the files on open5e. If I were to share the statblocks I save from the statblock generator, would that allow someone else to update the data?

augustjohnson:
This is deployed out to beta.
https://api-beta.open5e.com/v1/monsters/azer-forgemaster-a5e/

PR #336, opened 2023-10-12
Fix missing bonus actions

Sturlen:
Fixes the missing bonus actions part of issue #335 . Monsters now have a bonus_actions field which will be populated with their bonus actions, if any. Only for API v1.


augustjohnson:
Code looks fine.

augustjohnson (approved):

Issue #258, opened 2023-05-27
Switch importer over to use Fixtures instead of custom import


augustjohnson:
https://docs.djangoproject.com/en/dev/topics/db/fixtures/#fixtures-explanation

The idea here is that in our /data directory, the file content is django generated fixtures instead of custom json. This makes input and output much cleaner, and aligned. We could use django manage.py dumpdata to produce them and django manage.py loadddata to consume them and load the site.

https://docs.djangoproject.com/en/dev/ref/django-admin/#django-admin-dumpdata

augustjohnson:
Solved by #337

augustjohnson:
This is done in staging now.

PR #337, opened 2023-10-13
Api import adjust

augustjohnson:
!!! BIG CHANGE !!!

This PR will modify the directory structure so that data will be sorted into /data/v1 and /data/v2/, and all data fields are lined up correctly with their relevant models.

This now uses fixtures instead of custom import logic for data import.

Quicksetup can be used exactly like it was before.

I tested this by reviewing various models before and after and confirming the data appeared as expected.

I deprecated a LOT of logic (importer, populatedb, quickload).


augustjohnson:
This does break the /manifest endpoint.

augustjohnson:
This also addresses #258

augustjohnson:
FYI this needs a change because of bonus actions.
https://github.com/open5e/open5e-api/pull/336

augustjohnson:
I was definitely doing some wonky shit on this one.  I branched from the wrong place.  Closing. I'll resubmit something.

PR #339, opened 2023-10-18
Planes - Beyond the Material

nstefanski:
Adding the section "Beyond the Material" to the planes API with a separate entry for each header.

Exceptions: 
1. There was an extra superfluous heading for the Outer Planes. If you compare the SRD to the PHB, this was actually a table header that seems to have been left in by mistake. The text under this header belongs with the first Outer Planes header.
2. Demiplanes are listed under a new "uncategorized" section. Again, this is based on the PHB, where demiplanes are one of several headings under "Other Planes." Since the rest of these were removed from the SRD, demiplanes seem to be under the "Outer Planes" header, but should not be.

Also made some updates to add the parent field internally


augustjohnson:
Available here:
https://api-beta.open5e.com/v1/planes/

PR #360, opened 2023-10-25
fix: Remove Extra Mage Reaction

Sturlen:
Fix after bug report from Trick on the Discord:

> Need to remove the spell Misty Step from the reactions of the Mage in the Level Up A5E Monsterous Menagerie.
> Found on Page 482 of said book. 
https://discord.com/channels/470638316103008268/1166787857625657424


augustjohnson:
This has NOT been fixed as of this:
https://api-beta.open5e.com/v1/monsters/mage-a5e/
It should be a pretty simple fix though, look for mage-a5e in https://github.com/open5e/open5e-api/blob/staging/data/v1/menagerie/Monster.json

PR #361, opened 2023-10-25
Original Subclasses

nstefanski:
Creating a new pull for additional O5E original subclasses, I will add more to the same branch.


eepMoody:
Heyo @nstefanski. I'm working through these and getting them updated to the new data model, I think I got everything copied over correctly here, LMK if anything looks off!

eepMoody (approved):

PR #362, opened 2023-10-26
Additional Tal'Dorei material

nstefanski:
This commit finishes the majority of the open game content in the book. Two small sections not included are a list of contraband items and stats for a skyship. These are not really formatted like magic items but instead like regular equipment, which doesn't have a dedicated api, so I decided to exclude them.

Note that this book has some fairly large sections of optional rules. I put these in the sections api with parent set to "Character Options", the name of the chapter where they appear. However, if there are enough sources with optional rules sections, maybe there could be another section of "Optional Rules"?

The names of magic items and some monsters were altered to remove Product Identity.


augustjohnson (commented):
Merging this does not materially change the product.

augustjohnson (approved):

PR #376, opened 2023-11-06
Fix enchanter abilites markdown

Sturlen:
A parsing error in the markdown seems to have left some extra symbols which crash the md-viewer on the open5e site when viewing the [Enchanter](https://open5e.com/monsters/enchanter). The star needs to be double escaped since it's contained within JSON which contains Markdown to preserve it.

Reported on the [Discord](https://discord.com/channels/470638316103008268/1170789236086161468/1170789236086161468) by victronicon.


eepMoody:
Yo! I think this got splinched in the new data updates that were happening around the time you submitted it, sorry it took so long to get it resolved.

I think this merge correctly ports the fix over to the new data file?

eepMoody (approved):

PR #378, opened 2023-11-07
Approval Testing

Sturlen:
Replace old unit tests with approval tests for the public API. Currently tests all endpoints to make sure the structure does not change unless approved. This means you can confidently refactor code and make sure there are no unintended changes to the user experience. These test will be run in a workflow on all PRs.

This is only a starting point which can be expanded upon later to make it more comprehensive.


Sturlen:
Found some inconsistencies in my local version and github actions running the tests. tests should be rerun now.

augustjohnson (approved):

PR #357, opened 2023-10-25
Black flag SRD content

Sturlen:
First pass on parsing the Black Flag Reference Document. Currently only monsters. Parsing is not yet fully complete, so some monsters are quite broken. Needs proper license/document too.

Based on issue #338

PR #382, opened 2023-11-08
Pr357

augustjohnson:
Credit to @Sturlen and work in #357, which I lost the commit history of because of merge headaches.


augustjohnson:
@Sturlen this merges in the black flag data in the new format, but because of an import issue, it does NOT import it.

PR #384, opened 2023-11-10
Add *.recieved files to .gitignore

Sturlen:
Only the approved files should be commited.


eepMoody (approved):

augustjohnson (approved):
Looks good no concerns.

PR #387, opened 2023-11-13
Fix overwritten magicitems

Sturlen:
The slug field for magic items was not read, causing magic items with the same name from different publishers to overwrite each other. Using the slug field again adds back about 300 items which were previously not showing.

Thanks to @mshea for the detailed bug report on Discord.


augustjohnson:
Ahhhh this makes sense.  I will have to do some trickery to get this all converted over to my 1.4.0 release branch, but I'll get it.

augustjohnson (approved):

PR #391, opened 2023-12-04
Fixing a few poorly mapped races.

augustjohnson:


augustjohnson:
@Sturlen this should fix the issue.

augustjohnson:
Removing this because @Sturlen fixed it better in #392

PR #394, opened 2023-12-06
Cleanup docs

augustjohnson:
Updating some basic docs to represent the new data.
I also moved version.py out of the root, that's a bad place for it.


README.md
```diff
@@ -123,16 +123,15 @@ Smaller edits such as spelling mistakes can be edited directly in Github. For la
...
 
-## Adding a new sorce
+## Adding a new source
 
-To add a new source, create new directory inside `data` and a `document.json` file that credits the source and links to the license it was published under. An example of this can be found [here](/data/a5e_srd/document.json). You can then add a json file for each category of content. See an existing source, such as the 5.1 SRD to see how these should be structured.
+To add a new source, create new directory inside `data/v1` and a `Document.json` file that credits the source and links to the license it was published under. An example of this can be found [here](/data/v1/wotc-srd/Document.json). You can then add a json file for each Model Fixture of content. See an existing source, such as the 5.1 SRD to see how these should be structured.
```
Sturlen, line 130: would be great if you could elaborate on the "pk" field in a document. what should be put there and the fact that the "document" field in other models need to point to it.

README.md
```diff
@@ -147,11 +146,6 @@ Approval tests are run against the approved files in `api/tests/approved_files`
 
 Recieved files shall not be included in the git repo.
 
-# Deployment
-
-The API is normally deployed via [Docker](https://docs.docker.com/get-started/). You can either build and host it yourself, or use one of the tested providers below:
-
-
```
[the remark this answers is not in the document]
augustjohnson: Nice catch, that was less noticible editing in monospace.

README.md
augustjohnson: I added a bit more, but I still don't have a great, clean answer for v2 for making PKs when there's conflicts.
Acolyte is a good example. There's one in wotc and one in a5e.

Prepending the publisher name is good, but in what hierarchy? Which publisher is best?
Prepending the ruleset might work, but could there be two "Obsidian Greatsword" items both in 5e?

I suppose I'd tend towards publisher if I had to choose. Maybe we could add a publisher shortname?

Any thoughts?

Sturlen: The convention so far is to add the document slug as a suffix for all 3rd party publishers.
Eg. obsidian-sword is the 5e srd item and obsidian-sword-a5e is the a5e version.

eepMoody: I would also lean towards including a suffix.

As far as dupes, I think maybe we stick with the current convention (to avoid breaking links), then add the ruleset if it's different?

So if a5e publishes a version for D&D Next, then their item becomes `obsidian-sword-a5e-ddnext` (or whatever we decide on for that key?)

Since they're the original publisher of the "a5e" ruleset, we don't need to repeat it twice for them (same as we do with wotc-srd for the wotc-5e ruleset)

eepMoody (approved):

PR #395, opened 2023-12-06
Remove bonus action duplicates

### LINKED ISSUE #393: Monster reactions duplicated into bonus actions
Example found in the [a5e Malcubus](https://api.open5e.com/monsters/malcubus-a5e/):
```json
"bonus_actions": [
        {
            "name": "Shapeshift",
            "desc": "The malcubus magically changes its form to a Medium or Small humanoid or into its true form. It can fly only in its true form. While shapeshifted, its statistics are unchanged except for its size and speed. Its equipment is not transformed. It reverts to its true form if it dies."
        }
    ],
    "reactions": [
        {
            "name": "Living Shield",
            "desc": "When a creature the malcubus can see hits it with an attack, the malcubus can give an order to a creature charmed by it within 5 feet. The charmed creature uses its reaction, if available, to swap places with the malcubus. The attack hits the charmed creature instead of the malcubus."
        },
        {
            "name": "Shapeshift",
            "desc": "The malcubus magically changes its form to a Medium or Small humanoid or into its true form. It can fly only in its true form. While shapeshifted, its statistics are unchanged except for its size and speed. Its equipment is not transformed. It reverts to its true form if it dies."
        }
    ]
```
Also in the [a5e Lich](https://api.open5e.com/monsters/lich-a5e/)

The duplication appears to be in the source data of the [monsters.json](https://github.com/open5e/open5e-api/blob/89e87f68d9cf4d7b42b146d9bf8587c80e4b5371/data/v1/menagerie/Monster.json#L18491) file where the Malcubus is.

Since these are identical duplicates of bonus actions, it should be possible to create a script that simply removes these.

The source seems to be commit https://github.com/open5e/open5e-api/commit/a58a55b528d5c8b7e6fb1aefee429feb06863358 , when the monsters were first added.

Reported by Ankh on the [Discord](https://discord.com/channels/470638316103008268/1160455516191989830/1181757158048677969).

Sturlen:
Closes #393.


augustjohnson (approved):
I can't check the source, so I'm trusting you on the actual determination of what's a reaction/bonus action, but the change looks good to me.

PR #400, opened 2023-12-27
Bugfix: Move core classes from O5e to wotc-srd

eepMoody:
@augustjohnson want to get your eyes/review to make sure I did everything right in the new system.

Noticed that a bunch of the core classes were mis-labeled as "05e original", which I _think_ is related to the original subclasses that were being added in the old import system.

This should put them back in the correct document source (the 5e SRD)


augustjohnson (approved):

augustjohnson:
no issues here, i believe this got flubbed on my one-time export.

PR #401, opened 2023-12-29
added size rating for sorting

uribracha2611:
### Overview
This pull request addresses a bug in the sorting functionality by introducing the `size_rating` field. `size_rating` 0 is for tiny, 1 is for small, 2 is for medium, 3 is for large, 4 is for huge, 5 is for  gargantuan and  6 is for titanic
### Changes Made
- Added the `size_rating` field to the Django model.
- Populated existing data with appropriate `size_rating` values.

### Purpose
The primary purpose of this addition is to facilitate sorting based on the `size_rating` field.

### Related Bug Fix
This pull request resolves a bug related to sorting by size, identified as [Issue 399](https://github.com/open5e/open5e-api/issues/399). The bug caused incorrect sorting by size as described in 

### Local Development
On my local machine, the changes work as expected, and the sorting functionality is functional. However, I encountered issues with running the local tests. The local tests did not pass due to the following error
ImportError while loading conftest,  there is not a lot of information on that error, so i couldn't fix it

Thank you for your time and consideration.

PR #404, opened 2024-01-05
Added Nord Games Ultimate Bestiary Revenge of the Horde

Rrrrrrrrrrnt:
# Original:
### api.v1:
- added document
- added monsters
### api.v2
- added publisher

1. spell check and error checked, all monsters done.
2. checked local open5e-api, completed for all monsters individually, document and publisher.
3. checked local open5e website, completed for all monsters individually.

# 2023 Jan 14th update:
### in data/v1/UB-RotH/Document.json
- addressed indentation issue that may be causing a "check fail".
### added: 
- data/v2/nord-publishing/UB-RoTH/Document.json

### api/serializers.py
- 'mythic_actions',
- 'lair_desc',
- 'lair_actions',
### api/models/monster.py
- added field above and confirmed in local tests that they work on both `monster.json` files with (new) and without (old) them

## new attribute: side_id
- adde a new attribute called size_id to the `api/models/monster.py` and `api/serializers.py`
- this assigns a numerical number from 0 (Tiny) to 6 (Titanic) to each monster in the model to facilitate future migration to api.v2, this also links to PR #401 

Hopefully all tests will work this time and then we can get on with adding new sources :laughing:


eepMoody:
@Rrrrrrrrrrnt getting this checked out and ready to merge in! Was traveling this week for work and getting caught up

eepMoody:
Looks like we have a failing test related to /documents. I'm not quite parsing what's up with it (and can't get to my usual computer where I've got all this stuff set up right now) but hopefully it makes more sense to you? 😆

Rrrrrrrrrrnt:
do I need to include a document.json file under api v2?
I only have one under api v1.

Rrrrrrrrrrnt:
checked document in local environment, http://127.0.0.1:8000/v1/documents/

all working, so I reckon it is the fact that the document tis missing in v2, as mentioned above I'll include the updated in next PR

Screenshot 2024-01-14 at 20 00 59

Sturlen:
> new attribute: side_id

I like the idea of this, but I think it deserves it's own separate PR or be merged with #401. Splitting content and API changes will help to make reviewing and testing these changes easier.

Rrrrrrrrrrnt:
@Sturlen and @eepMoody 

you guys know best (I had not done anythink like this before i joined the Discord a month ago).

I can delete the `size_id` field and generate a new commit if that will make the reviewing esier.

#401 also has content and api change, a lot more content change than this one as all the monster.json files were edited.

eepMoody:
I think in this case we should be safe to do both at once, went through and ran them locally, and the changes all seem to be solid after testing them locally! Especially given that the API changes are being made specifically in support of this content.

But definitely agreed that as a general rule we should be doing them separately

uribracha2611:
i think i found out why it doesn't sort properly it is because the property decorator doesn't add the column to the database and then it can't sort by it. There's a [Stack Overflow question](https://stackoverflow.com/questions/17682567/how-to-add-a-calculated-field-to-a-django-model) about this topic.

augustjohnson:
It appears as though Nord Games noted that "creatures" are product identity, therefore cannot be reproduced in the site. This means that merging this would be a violation of their license. It will not be merged.

Issue #296, opened 2023-07-11
V2 classes


augustjohnson:
Use this as inspiration for the actual responses.
https://github.com/bradyslot/dndex/tree/master/src/data

augustjohnson:
He's moved on to use Rust instead, but here's an example json object for a class:
https://github.com/bradyslot/dndex/blob/fac806e7d66757540299a861c683d20284fe1f02/src/data/classes/warlock.json

augustjohnson:
#407 addresses this.

calumbell:
With #407 merged to staging this issue has been resolved. I am going to go ahead and close it

PR #407, opened 2024-01-23
V2 classes

augustjohnson:
This is the v2 classes, including SRD data. It's missing a couple of main features (proficiences, starting equipment), but I think those will come along shortly, I want to have a standard approach between classes, races, feats, etc, and I don't quite have that yet.


calumbell (changes requested):
I encountered a bug in the `CharacterClass` model in which crashed my local server. Line 55  of `api_v2/models/characterclass.py` was throwing an array out of bounds error.

```
53  def hit_points(self):
54    hit_dice_name = "1{} per {} level".format(self.hit_dice, self.name)
55    split_dice = self.hit_dice.split("d")[1] # <-- Here
      ...
```

Looks like the problem is that we are spliting the hit dice value; "D6", "D8", etc., by the character "d". Because the cases don't match ("d" and "D") it doesn't split the string into an array correctly, so when you index into it you hit an error.

Replacing "d" with "D" on line 54 fixed this for me:

```
53  def hit_points(self):
54    hit_dice_name = "1{} per {} level".format(self.hit_dice, self.name)
55    split_dice = self.hit_dice.split("D")[1]
      ...
```

calumbell:
This is looking great to me so far! Curious to hear your thoughts on a few points:

1) I noticed that a call to `/classes` endpoint returns both classes and sub-classes. On one hand this makes sense to me (if somebody is browsering O5e classes, they probably already know the WotC classes and are specifically looking for subclasses, why not surface these?). But on the other hand, subclasses aren't classes. My instinct would be to have subclasses at children on the class route (ie. `/paladin/oath-of-devotion`), so I am curious about the thought process here. You have spent a lot more time working with the API then I have so there is a good chance I have overlooked something!

2) Is there a simple way of sorting the properties returned by the API? Perhaps on the view or serializer? Currently the class/subclass name is right at the bottom of the query when it should probably be at the top.

augustjohnson:
Solid feedback.

1. I feel pretty strongly that we should consider classes and subclasses to be siblings at the data layer. This would be consistent with races/subraces as well. The way I see it is that if they're sibling rows in a table on the server (which they are) then, they should be presented as siblings on the API. I think there's lots of simple ways the UI layer can handle this, and it may even reduce some work downstream (why reimplement the same logic for class/subclass when it's the same dataset?) Basically, it boils down to: The API presents the DATA as it is structured, and I chose to structure these as both the exact same object type, with a nullable "parent" field.
2. I think so? https://www.django-rest-framework.org/api-guide/serializers/#specifying-which-fields-to-include may be the way to do it. It would likely make my relatively simple serializers much more verbose, and it doesn't really game much in usage down the road, so it's not a high priority on my side. That being said, I know it's weird to have calculated fields that aren't super important above Name when reviewing something.

calumbell (changes requested):
Awesome! That fixes it. Everything is running as it should on my local machine Using the `lower()` method is a good idea. Much more resilient!

While we have eyes on this PR, there are a couple small things we may as well nix.

1) [BUG] In the `hit_points_at_higher_levels` field, the average value is rendered with a trailing decimal point.

Screenshot 2024-03-06 at 09 05 16

2. [SUGGESTION / DISCUSSION POINT] The model names `Feature` and `FeatureItem` are a little ambiguous. When you see that they are defined in `models/characterclass.py` it is clear that they are class features, but if you miss that bit of context it is unclear precisely what types of features they contain (racial features? background features?, feats but with a longer name?). Would you be open to the suggestion that we change these model names to `ClassFeature` and `ClassFeatureItem`? This should improve the clarity of the code, and will avoid having multiple models of the same name when we begin implementing races and backgrounds for v2. If they are intended to be a generic Feature model, which covers  class/racial/background features, then I don't think the `models/characterclass.py` route is the best place to define them

augustjohnson:
I've made the following changes:
1. I changed the input data to use lowercase D for hit dice.
2. I've used floor division for the calculation of that number, which should work a bit better.

Regarding the class name discussion, that's probably a good idea. There's lots of implications though. I've created #423 for this work.

calumbell (approved):
Looks good! The trailing decimal is gone on my local branch.

Opening a new issue to investigate changing the Feature model names is a good idea if this decision impacts other parts of the codebase. This PR is already big enough!

I am going to go ahead and approve this. We will almost certainly want to tweak this endpoint in future, but for now everything is working to a point that satisfies the initial issue (#296). Nice work!

PR #411, opened 2024-02-05
Add Missing Sections

nstefanski:
A few rules sections from the WotC-SRD (particularly those that appear before list sections like "Magic Items" and "Monsters") were missing from the open5e data set, this PR adds those in.

Also noticed that some sections for which data exists (Spellcasting and two of the appendices) don't appear on the open5e site because their parents are not in the menu, so I re-categorized these following the example of the "Conditions" appendix which was already categorized under "Rules".

Note: it looks like the dataset includes automated timestamps. I wasn't sure if the data would cause problems without a timestamp, so I manually entered dates when I completed the data entry.


augustjohnson (approved):
Good work, this will definitely make the site a better place.

PR #415, opened 2024-02-24
Pass tests for: Updated TOB3 license and fixed Godslayer

Sturlen:
Fix to pass the tests. Includes the work done by mshea in PR #414 so this PR should be merged in it's stead.


calumbell (approved):
Nice work! Runs well on my local machine. Nothing explodes. Changes to `/monsters/godslayer` look good.

It is worth flagging that this PR fixes a *whole lot more* than the title lets on. Also included are changes to `data/v1/menagerie/Monster.json`, which add in missing skill modifiers to loads of the monsters from the Monstrous Menagerie. Two birds one stone, but these would probably be better served as two seperate PRs in future so that this poor reviewer doesn't get too confused!

Thanks again for your contribution! I'll go ahead and get this PR merged

(In the course of reviewing this PR a bug in the front-end to light surfaced: skills are not being rendering correctly on the `/monsters/[id]` route, even when they are being returned correctly by the API. I'll go ahead and open an issue there)

PR #417, opened 2024-03-01
V2 spells

augustjohnson:
This PR includes the model and data infrastructure for v2 spells. It includes many changes. I'll try not to document them all here, but the basics are:

All fields from v1 are available in a clean data form (for example, no more "yes" for a boolean field). Many have been slightly renamed for clarity.

Various new fields have been added, such as attack_roll (Whether or not the spell requires an attack roll), target_count (the number of targets of the spell), and casting_options, which is a list of deltas to the intial spell instance that each represents a different way to cast the spell.  For example Ritual, or Upcasting Magic Missile at level 3 for 5 darts.

I feel very confident (maybe 90%) about the data quality within the wizards SRD. I feel less confident about the data from the other data sources, but I have converted it all.

Known issues [Data related]:
In some cases there are two conflicting casting options.
In some (all?) cases, the ritual casting option is missing.

I will fix these in follow up PRs.


augustjohnson:
These last few commits are purely just data that I needed to enter values for.  For example, Bestow Curse doesn't follow an easily automatable higher-levels format.

calumbell:
Nice one, thanks for all the work you have put into this PR.

There is a lot to sift through here, it will take me a minute to fully wrap my head around everything. I was wondering if I could ask a few clarifying questions to make sure I understand what is going on here?

1) I see that the `casting_options` array returned by the API contains upcasting options for each spell. While helpful, this array contains a lot of duplicate/redundant information. This is especially obvious when you look at cantrip damage scaling: for example, on acid splash (`/v2/spells/acid-splash/`) the `casting_options` array contains 21 items, only 4 of which are actually unique options (the level thresholds at which the damage changes). Wouldn't it be more efficient to only include the character/spell-slot level thresholds at which the spell effect changes?

2) On a similar note, the first option of the `casting_options` array for each spell is always a `default`, here all of the fields are null. My instinct is to lose this as it is increasing the size of the API response and not providing any information, is there any pros that I am overlooking here?

augustjohnson:
**Casting Options design**
So the basic idea I had here was to provide any client with a comprehensive list of the different ways a spell could be cast, and a platform to describe the effect (which changes, based on the casting option selected).

For a spell like Fireball, this is intuitive. You cast it at 3, and it does 8d6. You cast it at 4, it does 9d6. So the casting options just have an input of slot level and an output of damage. But for other spells it gets much more complex.

Magic Missile is a good example, it doesn't scale damage directly, it actually scales target count. Some spells are not well represented in scaling yet, like the summons, but that can be a longer term problem. Ritual casting is another example, where you're FORCED to the base casting effects, but you do not consume a slot, however the casting time is increased.

So to support both Fireball and Magic Missile, and Ritual tagged spells (and to genericize this to meet 90%+ of scaling spells), I need to have a set of fields defined in the casting option. That's how I defined the fields that exist in CastingOptions.

Regarding whether or not it's best to actually present the fields that are null, in my  opinion it's always better to be explict. "The duration is unaffected by this casting option" is a better statement that omitting the duration from the response, especially, when you have strangely scaling spells like this:
https://open5e.com/spells/conjure-fey. 

So that model works well for spells that require slots, but Cantrips do not require a slot, they scale on player level. I used the logic "better to be explicit" again here, and to use the same general method I used on the others, identifying the player level by the "type" field. Could it be simplified by some sort of range field? Possibly. I viewed the tradeoff in implementation complexity (on both server and client sides) as too high to introduce a new way of scaling.

Also, in the end, it's CODE reading these verbose objects, not people, so it doesn't really matter that they're verbose.

augustjohnson:
**Default Casting Option design**
I chose to include a DEFAULT casting option on all spells that could be cast. In my mind, the frontend logic can now calculate fairly simply a few items like "how many different ways can this spell be cast?" - For Sanctuary, 1 (default). For Find Familiar, 2 (default, ritual). Given more than 1, force a CHOICE to see the results. That seems intuitive to me. I would generally agree though, that default will ALWAYS be nulls for all supported fields.

For cantrips though, I could see a solid argument being made for deletion of "default" though. Do you have a preference on that?

calumbell:
Thanks for taking the time to explain your design to me. I can definitely see why structuring the data can be helpful in lots of contexts, especially in something like a character generator / virtual character sheet.

One option that I completely overlooked when writing my last comment was using query parameters in fetch URL on the front-end to request to only request the fields that you needed for a specific context. I know that you are able specify which fields you'd like Django to return by using the `field` query param (ie. `.../v2/spells/?fields=name,level,school`), but do you know if it is possible to exclude certain fields in a similar way?

Filtering by field is currently not mentioned on the [Open5e API page](https://open5e.com/api-docs), definitely a good issue to open on the front end repo.

With all this in mind, i'll take another look at the PR and see how everthing else look.

augustjohnson:
The ?fields param will remain supported. We don't have an "exclude" param. It's not quite graphql :).

calumbell (approved):
Armed with a little more context about the design process behind version 2 of the API, it looks like this PR satisfies the goals of the initial issue. Still plenty of features left to add, but great work so far.

Everything works as expected on local, happy to approve this

PR #421, opened 2024-03-02
added release workflow

coderatul:
don't know whether it would work or not because to check build has to run whihc needs docker credentials

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,46 @@
+name: Create Release
+
+on:
+  push:
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+
+jobs:
+  build_and_release:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


augustjohnson:
Ok, looking in the right direction.
I don't think "install dependencies" is a necessary step, because we don't even use pip, and we're not packaging up a wheel, so you can remove that.

Can you change the zip of the codebase name to be something more like "open5e-api-1.1.1.zip" or something?

I also think DRAFTING the release is probably a good idea, so I think the process would roughly be:
Tag a merge
which triggers this.
Approve a DRAFT after reviewing.

augustjohnson:
Ah, sorry, remote the build step as well, I'm imagining literally just code (for now) is part of our release.  There's no way at the moment to really use open5e as a dependency.

coderatul (commented):
any review ? any changes to be made

augustjohnson (approved):

Issue #344, opened 2023-10-21
V2 Vehicles


augustjohnson:
Vehicles endpoint, model, etc.

augustjohnson:
Not yet. this will be an extension of object or maybe item I think.

augustjohnson:
Ok, so my determination on this is:
1. Since just about any creature can theoretically become a mount, there's no new TABLE for that, but I will be creating a creatureset for "common-mounts" that includes the list in the SRD.
2. For vehicles, the list is literally 6 total "waterborne vehicles" and the only new field they have (from Item) is Speed, which is listed in mph rather than any other common unit. My plan is to simply add an item entry, and have a description of the speed as part of the item description, as well as putting them into a waterborne-vehicles itemset.  It loses the easily parseable speed column, but it should be good enough.  If there's a long term space based dataset or something where this matters, then we can adjust later, but for now it should be ok.

augustjohnson:
Merged as part of #422

PR #422, opened 2024-03-03
V2 vehicles

augustjohnson:
This should address #344.  Just data entry at this point, adding creatures and items to sets.


calumbell (changes requested):
Looks like a good solution to an unusual data-set. Nice work!

One thing that springs to mind is that the `category` field is handled inconsistantly across creatures and items.

In `CreatureSet` the `common-mounts` entry contains an a field called `creatures` which contains the primary keys of creatures considered common mounts. To set a creature as a common mount is to go into the CreatureSet and update this array. (The new creatures have a field called `category`, which is currently set to "Miscellaneous")

In `ItemCategory` the `drawn-vehicle` and `waterborne vehicle` this array is absent, instead the category field of an `Item` contains the pk of an `ItemCategory`. To add an Item to a ItemCategory, you go to the Item and update the `category` field.

Is it be possible to change Creature/CreatureSet so it confirms to the same structure as ItemCategory? I could well be missing some context as to why things are set up this way, so please do correct me if I have gotten this wrong. Only flagging this because it was something that confused me during my review, and it could well confuse other developers working on this feature down the road.

Tangent over. Everything is working perfectly on local!

augustjohnson:
The word "Category" is used with many different purposes in the SRD. For example [Light, Medium, Heavy] are listed as categories of armor, [Small,Medium,Large] are listed as Size Categories, etc.

What if... any Object (which both Creature and Item inherit from), has a category? And this would map to:
[Monster, Miscellaneous Creatures, Nonplayer Characters] based on the appendix they are in for the SRD and
[Armor, Jewelry, Gem, Tools, etc.] for items.  I think that's a pretty good idea actually.  Maybe I can refactor ItemCategory to be ObjectCategory and have all of this.

I have logged #424 for this work.

PR #425, opened 2024-03-15
V2 expose enums

augustjohnson:
Exposing a few utility classes, and the enums.


calumbell:
Hi, thanks again for your work!

This is a pretty substantial PR which touches on several existing issues, but it doesn't include much context on what has been changed. Best I can tell, this looks like a refactor of several data models to redefine certains fields as enums. Is that the basic shape of it?

Also, it looks like something has happened to the spell schools in the v2 API data. They all read as being from the school of Evocation?

augustjohnson:
Eh sorry about that, that wasn't even up to my (admittedly low) standards of PR documentation.

The goal of this PR was overall, to expose a set of values for well-known and re-used fields.  I also wanted to consolidate hardcoded values to a single place. Some of these remain hardcoded (but now centralized), and some I have written into custom models.

Size is the best example of a custom model.  Size was being stored as an integer before, but within the scope of this PR was to make a clear "size" that was shared as a field across various models (object, creature).  In that particular case, my solution was to create (and expose) a full object and add some fields.  Similarly, I adjusted Spell Schools, and Damage Types and Item Rarities. As a note, there is not a Rank field on both Item Rarities and Size that can be used to sort.

But a few other items I just centralized into a hardcoded values section (called enums). Now there's also a /enums endpoint, which exposes some of the "hardcoded" values used, like Create Attack Types (Weapon, or Spell), and Spell Target Type Choices, (creature, object, point, area).

Functionally, this has very little changes on the endpoints/models that are existing in v2, even though those files were edited.

Regarding all spells being evocation, yes, that's a bug of some sort. I'll fix it.

augustjohnson:
Fixed.  Did you know that a5e has a distinct spell school called "Transformation"? I didn't! But luckily the app was forgiving enough to clearly show me the problem, and to allow me a simple solution.

calumbell (approved):
Awesome. Thanks for taking the time to explain what is going on in this PR, it was very helpful! Refactoring these hard-coded/enums values so that there are all in one spot seems like a good design choice to me.

I have had a poke around the source code and nothing strange jumps out at me. Tested the PR out on local and everything appears to run as expected. Good work!

PR #426, opened 2024-03-28
Adding tob1 2024 monsters

mshea:
Sorry, this branch is mislabeled as TOB 3 but it's actually TOB 1 2023 monsters. I spent a bunch of time parsing TOB 1 2023 monsters and I've checked them pretty thoroughly. I think they're ready to merge.


calumbell:
Hi, thanks for your work getting these monsters input and formatted! There really is quite a lot here!

I encountered a few problems when I tried to build the database my local branch. I have fixed them to the point that the server builds successfully,  but isn't passing all the tests yet (will look into this tomorrow). I have pushed my changes to this PR.

1) Certains fields in the Monster model (`bonus_actions_json`, `reactions_json`, `legendary_desc`, `legendary_actions_json`) were not nullable, so setting their value to `null` in `Monsters.json` was causing Django to choke. 
This was fixed by updating the Monster modal to accept null values for these, which is probably how it should have worked to start with (other sources handle the null value as either an empty string or `"null"` as a string, which seems weird to me)

2) The `document` field in the Monster.json file is supposed to contain the primary key (pk) of that monster's source document. For TOB-2023, that value is 47 (as defined in `data/v1/tob-2023/Document.json`). I fixed this by updating the value of the `document` field for all the monsters you've added.

augustjohnson (approved):

augustjohnson:
New fields in the data, but not yet exposed in the API... we should make sure we're going in open-eyed to that change.  I'm considering that out-of-scope at the moment.

PR #427, opened 2024-03-28
Adjusting the document stats field.

augustjohnson:
The v2/documents path was erroring because stats were incorrectly calculated.


calumbell (approved):
Hi nice work!

I was able to recreate the bug by hitting the `v2/documents` when running the `staging` branch on local. When I load up the PR and run it on local, the endpoint now runs. I'll go ahead and approve this.

PR #428, opened 2024-03-29
V2 search

augustjohnson:
This is a DRAFT search, meant to replace the v1 search functionality and present equal or better results. First off, it is not available and not loaded by default, so to test this PR you will have to load it. In a future PR the loading will occur.

Method to load data into the search index:
python manage.py index_v1
(You can run that right after quicksetup, and it should work just fine).

Example searches to try:
http://localhost:8000/v2/search/?query=amulet&schema=v1
http://localhost:8000/v2/search/?query=Plate&schema=v1
http://localhost:8000/v2/search/?query=goblin&schema=v1
http://localhost:8000/v2/search/?query=magic&schema=v1
http://localhost:8000/v2/search/?query=asdf&schema=v1

Comparisons:
https://api-beta.open5e.com/v1/search/?text=amulet
https://api-beta.open5e.com/v1/search/?text=Plate
https://api-beta.open5e.com/v1/search/?text=goblin
https://api-beta.open5e.com/v1/search/?text=magic
https://api-beta.open5e.com/v1/search/?text=asdf

I do NOT have weights in the query yet, those will be added. I also think that the actual input filters/query is not really documented within the scope of DRF, because I had to manually build it. I am not sure exactly how to resolve that the best way. It doesn't seem to be cleanly possible to implement and STILL use FTS. Either way, this should be functional.


augustjohnson:
Using my admin privs to merge bypassing branch protections :/. For the sake of PROGRESS!

Issue #363, opened 2023-10-28
V2 Creature Sets


augustjohnson:
I'm working on this in the /creature_set branch.  I think I have the views all ok so it's data entry at this point.

augustjohnson:
I have merged this into release/1.4.0

augustjohnson:
- [ ] Merged in 1.4.0

augustjohnson:
Ok creature types is done and in staging at: https://api-beta.open5e.com/v2/creatures/

Creature set is not done.  There's a relatively small list of items in creature set, but they are important.  A Quasit is definitely creature type = fiend, but is also part of the creatureset demon, and shapechanger. Those two do not yet have a representation.

augustjohnson:
Creature Sets implemented and exposed. Data isn't really there yet, but that will be addressed as creatures get added.

PR #430, opened 2024-03-30
V2 creatureset

augustjohnson:
Finalizes CreatureSet which is #363


augustjohnson:
Using my admin privs to bypass branch protection.

PR #431, opened 2024-04-04
Minor docs/env improvements

Kordishal:
Changes:

- adds two lines to the gitignore for PyCharm and pyenv dev environment files and folders.
- fixes the pipenv installation link. Current one leads to 404 which lead me to the main page where installation is done via apt package manager, this again caused issues as it was an older incompatible version. Potentially could expand here to make it clear that pipenv needs to be installed with pip.
- Rename Modules to Dependencies as that is what you do there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,9 @@ server/whoosh_index/
 
 .vscode/
 
+# pycharm config
+.idea/
+# pyenv env file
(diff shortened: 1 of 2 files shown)


augustjohnson (approved):
No issues with these changes, merging.

PR #434, opened 2024-04-05
Minor docs improvement

Kordishal:
add pipenv run to each of the commands in docs/v2/data.md. This ensures that the .env file is loaded which is required as it asks for the SECRET_KEY env variable.

Remove trailing / on the export command as it is not required and just doubles up the slash.


augustjohnson (approved):
This is a more clear way to communicate the commands needed than the one I had been using.

Issue #405, opened 2024-01-06
Add a data entry guide.


augustjohnson:
We need a guide for new users to be able to add new data.

Rrrrrrrrrrnt:
Important steps for me were:

1. finding out where to save the files (e.g. v1 or v2)
2. finding out what file to use as a "template" (e.g. /data/v1/tob/monster.json)
3. understanding the correct syntax for each section (a good sample file will cover this one)
4. understanding how to test the API (installing Python, brew, Jango, pip, pipx, creating the environment 
5. understanding how to test the UI (edit /nuxt.config.ts to point to the local environment, install npm and run the environment)
6. reading the script in VS Code is messy and can be confusing when checking the work, for this I created a template in excel ([monsterjson rev2.xlsx](https://github.com/open5e/open5e-api/files/13851806/monsterjson.rev2.xlsx)) to help me add information, so it is easy to read when I am copying from the pdf.

The tech part was hard to do alone, as the guides online are incomplete or outdated.

All the above is local, then it came the part where i had to interact with everyone who knows what they are doing - and that was embarrassing (a little).

Took me a good couple of goes to get that I cannot click "commit" on VS Code just to same my work online.

The Commit part is daunting for new people (it was for me), as it is when you thing - "Oh I could be screwing the actual API or UI up!"

This is the first time I "coded" anything, I took it on as a challenge and a puzzle and I enjoyed - but it is daunting.

A very important note: everyone was and is super nice, very patient and helped me a lot - that is why I'm motivated to keep going and add another source to the collection (I am assuming the first one will be approved and will work fine).

Thank you for having me and I hope I can help more :)

augustjohnson:
Various document including https://github.com/open5e/open5e-api/blob/staging/docs/v2/data.md have been updated to include data entry. I believe this is resolved.

PR #439, opened 2024-04-08
Add a contribution guide

### LINKED ISSUE #243: Fill out community standards.
https://github.com/open5e/open5e-api/community - has some good advice.

Kordishal:
This pull request adds a contribution guide and improves the related documentation files.

This is a step to resolve #243 as the contribution guide will resolve one of the recommended steps.

And it is a first step to address #405. It does not go as deep as that issue requests, but should give some additional pointers. Especially when it comes to creating pull requests and how to approach contributions.

It is definitely not complete and can be expanded in the future, but I think this gives a good baseline. 

Additionally I have made some assumptions for how things should be done. So make sure to review if this is acceptable for you guys, as I used `we` which does not include me.

I moved and edited parts of the data docs file to make it easier to link.

This pull request only edits documentation files.


augustjohnson (commented):
Good reorganization overall.  Can you fix the django documentation links to be the current django version? We'll have to update this over time, but that'll be ok.

augustjohnson (approved):

PR #441, opened 2024-04-10
V2 search improved

augustjohnson:
These are expansions of capabilities introduced in the first search. This should allow for a nice v1 data compatible search endpoint.


augustjohnson:
Merging to get new search in. It should not cause issues with beta env.

PR #443, opened 2024-04-12
Tob3 hotfix

augustjohnson:
Fix for some new monster fields.


augustjohnson:
Hotfixing because of a bug in production related to the reactions json fields (and other similar fields).

PR #444, opened 2024-04-12
We only use 4 character "desc" for that field.

augustjohnson:


augustjohnson:
Merging in a bugfix with incorrect json.

PR #447, opened 2024-04-19
Use api.open5e.com for inline links instead of base_url

### LINKED ISSUE #446: Spell markdown breaks due to base_url placeholders
Spells like [Arcane Sight](https://open5e.com/spells/arcane-sight), [Snowblind Stare](https://open5e.com/spells/snowblind-stare) on open5e.com don't render properly because the markdown data is broken. The [As you were](https://open5e.com/spells/as-you-were) spell also has an inline link and still loads, however the link does not work.

The reason for this is that the markdown contains two links with `{{base_url}}` placeholders. Because they each include an underscore, the markdown compiler interprets half of both link as being a `<em>` element and this breaks the markup. Resulting HTML output:
```
[true seeing]({{ base<em>url }}/spells/true-seeing) and [detect magic]({{ base</em>url }}/spells/detect-magic)
```
From deep magic [spells.json](https://github.com/open5e/open5e-api/blob/e760416f345e0957011dc69ea17dc30d537b7de1/data/deep_magic/spells.json#L426).

There root problem here is not using a valid URL in the inline links. I think using the full https://open5e.com/api url as a base would be a better solution. For cases where you'd want to rewrite the link you'll still be able to do a search and replace, Not sure if this change is worth changing for v1 or if it should be addressed in v2.

---
Additional info:
Base urls were added in #287

Full error log I used to debug the problem:
```
[Vue warn]: [@vue/server-renderer] Template compilation error: Error parsing JavaScript expression: Unterminated regular expression. (1:7)
1  |  <div><p>The recipient of this spell gains the benefits of both [true seeing]({{ base<em>url }}/spells/true-seeing) and [detect magic]({{ base</em>url }}/spells/detect-magic) until the spell ends, and also knows the name and effect of every spell he or she witnesses during the spell’s duration.</p></div>
   |                                                                                                                                           ^
[nuxt] [request error] [unhandled] [500] Invalid regular expression: missing /
  at Function (<anonymous>)
  at ssrCompile (F:\Projects\open5e\node_modules\@vue\server-renderer\dist\server-renderer.cjs.js:587:35)
  at renderComponentSubTree (F:\Projects\open5e\node_modules\@vue\server-renderer\dist\server-renderer.cjs.js:656:24)
  at renderComponentVNode (F:\Projects\open5e\node_modules\@vue\server-renderer\dist\server-renderer.cjs.js:638:12)
  at renderVNode (F:\Projects\open5e\node_modules\@vue\server-renderer\dist\server-renderer.cjs.js:752:14)
  at renderComponentSubTree (F:\Projects\open5e\node_modules\@vue\server-renderer\dist\server-renderer.cjs.js:707:7)
  at renderComponentVNode (F:\Projects\open5e\node_modules\@vue\server-renderer\dist\server-renderer.cjs.js:638:12)
  at Module.ssrRenderComponent (F:\Projects\open5e\node_modules\@vue\server-renderer\dist\server-renderer.cjs.js:84:10)
  at _sfc_ssrRender (F:\Projects\open5e\components\MdViewer.vue:66:31)
  at renderComponentSubTree (F:\Projects\open5e\node_modules\@vue\server-renderer\dist\server-renderer.cjs.js:692:9)
```

Sturlen:
Closes #446. This fixes the markdown rending issues mentioned in the issue. This could cause issues for applications excepting the base_url jinja template, but I'm not sure how widely used this is. It is also necessary for fixing inline links on open5e.com. 

This would affect both v1 and v2.


augustjohnson:
Good enough, I don't love the hardlink to open5e in the in the data, but i certainly don't have a simpler solution.

augustjohnson (approved):
Changes make sense.

PR #459, opened 2024-05-24
V2 keys refresh

augustjohnson:
Some basic renaming, and a checking file.


augustjohnson (commented):
just changing keys to be consistent with newer standards.

PR #460, opened 2024-05-26
Change description to desc in tob-2023

Thomvis:
Tob-2023 is the only document (as far as I have encountered) that uses "description" instead of "desc".


calumbell (approved):
Thanks for getting this fixed! It has been an annoyance working on the front-end site with the property names `desc` and `description` not matching.

The PR runs as expected on local. Checked perhaps twenty or so monsters with different combinations of action types. All looks good!

I'll go ahead and approve and merge to staging. Thanks again for your work!

PR #465, opened 2024-06-01
429 v2 environments

augustjohnson:
Implementing v2 environments. There's no mapping yet to creatures, but the environments themselves are there.


augustjohnson:
Adding in environments. Should be good to go.

PR #477, opened 2024-07-04
Fix Deep Magic spell tables

crhallberg:
The markdown the spell description for Deep Magic didn't have enough dashes to be rendered into a table.

Before:
image

After:
image


calumbell (approved):
Hey, thanks for your contribution! Worked great when I spun it up on local with a fork of the front-end pointed at it. 

I did  encounter some weirdness with tables included in the the `higher_level` field not rendering properly. This turned out to be purely an issue with the front-end repo, which doesn't currently parse the markdown in the Higher Level section as HTML, and instead just slaps the API response straight into the template. Thanks for surfacing this, will open this as a new issue over there

PR #481, opened 2024-07-17
Query Monsters by size and hit points

Sturlen:
These fields were not searchable. This created a problem for the new [Pagination](https://github.com/open5e/open5e/pull/530) feature for the Open5e site, which now relies on the API for filtering.


calumbell (approved):
Looks good, works great on local! Query params on the endpoints `/monsters/?ordering=size`,  `/monsters/?ordering=-size` ,`/monsters/?ordering=hit_points`, `/monsters/?ordering=-hit_points` return sorted results as expected. Nothing breaks and tests pass.

One small weirdness is that ordering by size doesn’t actually order by how large a monster is (tiny → small → medium → large) but by how the names of these size catagories sort alphabetically (garganutan → … → titanic). I believe this issue is already being addressed in API V2

PR #496, opened 2024-08-03
add default content renderers

eepMoody:
This resolves a common error when requests for HTML or XML are sent to the server, and should allow it to properly return one of our supported types in response.


eepMoody:
@augustjohnson want to take a quick look over this one to make sure I didn't miss anything? I believe this will resolve our most-common classes of errors on NewRelic and make it a lot easier to spot actual issues when they crop up.

augustjohnson (approved):

PR #497, opened 2024-08-03
AI Code review action

eepMoody:


eepMoody:
I think this is all good! It will catch actual errors in the code and suggest fixes, but avoids commenting otherwise. Seems to be fairly responsive.

PR #505, opened 2024-08-05
Expanded type.

augustjohnson:


augustjohnson:
So I've removed some of the errant deep serializers that were embedded in some of these objects, with the basic idea that a combination of fields=... and depth=1 should mostly do the trick.

augustjohnson:
Adjusting display methods.

PR #509, opened 2024-08-06
Fixied shadow elves spells.

augustjohnson:


augustjohnson:
Adding some simple data errors.

PR #520, opened 2024-08-22
Bugfix: V2 Documents - filtering by query param

### LINKED ISSUE #518: Bug: `v2/documents` endpoint not filterable by URL param
## Bug Description

The API V2 documents endpoint cannot filter its results by field in the same way that V1 could (via the `?fields` query parameter). This feature was utilised across the Open5e website to greatly reduce the amount of data requested from the API by the source selector, which really only required a source's name, publisher, and primary key.

## Reproduction

1. Checkout the `staging` branch and spin up a test server
2. Visit the endpoint `v1/documents/?fields=title,slug`, note how only the specified fields are returned.
3. Visit the same endpoint on API V2: `/v2/documents/?fields=name,key`, notice how results are not filtered by field.

## Context
Digging into the source I noticed the following line commented out in `API_v2/views/document.py` (line 28):

```
class DocumentViewSet(viewsets.ReadOnlyModelViewSet):
    """
    list: API endpoint for returning a list of documents.
    retrieve: API endpoint for returning a particular document.
    """
    queryset = models.Document.objects.all().order_by('pk')
    serializer_class = serializers.DocumentSerializer
    #filterset_fields = '__all__' <--- THIS ONE
```

It looks like this field tells Django REST that this endpoint should be filterable by all fields. When this line is uncommented however, the `v2/documents` route throws the following error when accessed:

Screenshot 2024-08-13 at 13 43 16

Which looks to be like Django REST is having a difficult time figuring what to do with a JSONField when it is filtered. The warning contains a link to the [Django Documentation for customising filter generation with overrides](https://django-filter.readthedocs.io/en/main/ref/filterset.html#customise-filter-generation-with-filter-overrides), which might be a good place to start!

calumbell:
This PR closes #518 by:

- Updating the V2 Documents serializer to inherit from `GameContentSerializer`, where the filtering by query parameter logic is implemented
- Add a filter override to the Document view so that Django knows how to deal with JSONFields

Tests are passing and all appears to work well on local: `http://localhost:8000/v2/documents/?fields=key` returns only keys, `http://localhost:8000/v2/documents/?fields=key,name`  returns only keys and name, &c

### DIFF
--- a/api_v2/serializers/document.py
+++ b/api_v2/serializers/document.py
@@ -1,5 +1,6 @@
 """Serializers for Ruleset, License, Publisher, and Document models."""
 from rest_framework import serializers
+from .abstracts import GameContentSerializer
 
 from api_v2 import models
 
@@ -27,10 +28,9 @@ class PublisherSerializer(serializers.HyperlinkedModelSerializer):
         fields = '__all__'
 
 
-class DocumentSerializer(serializers.HyperlinkedModelSerializer):
+class DocumentSerializer(GameContentSerializer):
     key = serializers.ReadOnlyField()
-    stats = serializers.ReadOnlyField()
 
     class Meta:
         model = models.Document
-        fields = "__all__"
+        fields = '__all__'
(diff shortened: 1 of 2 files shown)


eepMoody (approved):
LGTM! Tested locally and appears to be working

PR #521, opened 2024-08-25
expose ruleset & publisher name/key on documents endpoint

eepMoody:
The most common use-case for documents in practice is selecting content, which is often selected by publisher (expected: ruleset once it is available)

Given that common use-case and the small size of these resources, it seems more user-friendly to expose these properties on the documents endpoint instead of requiring a 2n additional server calls (two per document) to retrieve them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ server/whoosh_index/
 .python-version
 
 api/tests/approved_files/*.recieved.*
+*.sqlite3-journal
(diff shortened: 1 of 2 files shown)


augustjohnson (commented):
I think there's an additional piece missing here. The ability to filter on these values from each individual object. I'm going to take a crack at it before I approve this, but I'll use your branch.

augustjohnson:
Yeah I think that this is totally unnecessary. You can get the relevant data (minus the key, which is superfluous and is part of the url field) by doing this:
https://api-beta.open5e.com/v2/documents/?depth=2. I vote to close this without merging.

augustjohnson:
Closing without merge, the strategy has changed.

PR #535, opened 2024-09-15
V2 Feature: `/search` now returns more helpful data

calumbell:
This PR was created to address minor errorsin how the Open5e API V2 `/search` endpoint functioned which were exposed during the front-end migration over to V2.

These changes update which fields are included in the `object` properties of results returned by the `/search` endpoint. In some cases these were  changed to improve aesthetics and navigation on the front-end (ie. for a creature, it is much more important to sees its Type on the /search page, rather than, say, its AC), but in other cases these extra fields are vital to link to the correct page on the website (ie. the CharacterClass endpoint now returns both base- and sub- classes, these resources are accessed via different URL structures, so it is critical to have this information on the FE if we want to link to the searched resource)

### DIFF
--- a/api_v2/models/characterclass.py
+++ b/api_v2/models/characterclass.py
@@ -113,4 +113,12 @@ class CharacterClass(HasName, FromDocument):
 
         return text
 
+    def search_result_extra_fields(self):
+        return {
+            "subclass_of": { 
+                "name": self.subclass_of.name,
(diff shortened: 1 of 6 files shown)


augustjohnson (approved):

augustjohnson:
I like it, nice additions.

Issue #522, opened 2024-08-25
v2 Spells cannot be filtered


eepMoody:
These do not appear to have any exposed filters, we should probably make almost all available

augustjohnson:
I believe this is resolve now with @tylereed's commit.

PR #536, opened 2024-09-16
Bringing v2 spell filtering into line with v1

tylereed:
Added missing filters that were in v1 but not v2 for spells.

Some things that I noticed.

Classes use key for filtering, but I used name in School.  I can change this to be 'key' if that's our preferred method.

Classes entry is empty for all spells.  Looks like there is test code in `v1_spelllist_to_v2_spell_class.py`.  But I wasn't able to test if class filtering was working.

Filtering on casting_time is working, but it seems that the data is bad.  9 of the spells have a time of 'reaction', while the rest have a time of 'action'.  It seems to me that the converter function `get_casting_time` isn't converting from v1 to v2 properly.  This is something I can look into.

#522


calumbell (changes requested):
Nice. Changes appears to work well on local and tests are passing. Thanks for your contribution!

Just wanted to touch on a the points you raised in your PR before moving ahead:

**TLDR - requesting changes: allow spell school filtering on either`name` or `key`**

---

1) *"Classes use key for filtering, but I used name in School. I can change this to be 'key' if that's our preferred method."*

As someone who works with the front-end, filtering on `name` is nicer, it makes the query parameters much easier to read. But there is an argument for filtering using `key` - what if spell schools changed dramatically between different publications/rulesets and we needed to handle different versions in our data model?

I can't think of a reason *why* a 5e publisher would want to do that, but it could happen. If two different sources both had an Enchantment school of magic, but they were functionally different enough that we would need to distinguish between them, then filtering by `key` would be required.

Solution: allow both. Allow filtering by `school__name` and `school__key`.

---

2) *Classes entry is empty for all spells. Looks like there is test code in v1_spelllist_to_v2_spell_class.py. But I wasn't able to test if class filtering was working.*

This is the case on the live site: https://api.open5e.com/v2/spells/srd_fireball/ doesn't even include a class field, but it doesn't appear to be the case on the open5e-api `staging` branch. Might be worth merging from `staging` on your dev branch, spinning up a test server, and hitting http://localhost:8000/v2/spells/srd_fireball/?fields=name,classes to see if you have any luck. Here is what I see:

Screenshot 2024-09-22 at 11 55 25

This won't work with any A5e soells. The value of the `classes` property on all the Level Up Advance 5e spells is is currently a blank array, I think because the A5e classes haven't been added to the V2 dataset yet so there is nothing to link to yet.

---

3) *Filtering on casting_time is working, but it seems that the data is bad. 9 of the spells have a time of 'reaction', while the rest have a time of 'action'. It seems to me that the converter function get_casting_time isn't converting from v1 to v2 properly. This is something I can look into.*

Yeah, there is definitely something weird going on here , great spot! Misty Step, which should be a bonus action to cast, is saying it takes an action on both the [live site](https://api.open5e.com/v2/spells/srd_misty-step/?fields=name,casting_time) and development server spun up from the `staging` branch.

Screenshot 2024-09-22 at 12 03 15

I have only taken a precursory look at the conversion scripts, but your instinct looks good to me! If you felt up to looking into this feel free to open a new issue!

*Edited for formatting*

calumbell (approved):
I ended up implementing the changes suggested on my local branch, where I was using it to test the front end. Figured I may as well push it here to save us a little work.

Thanks for your contribution! Looks good to me, runs well. Does what it says on the tin

Issue #544, opened 2024-09-25
API V2: Rules endpoint


calumbell:
Currently there is no equivilent to the `/sections` endpoint point from API V1, which was used to get sundry rules that didn't fall under other endpoints.

When this feature was discussed over discord Discord, there was a consensus that the name `/sections` is not a great name for this endpoint, being both vague and the name of a HTML element. I propose that we use the `/rules` endpoint.

The work that needs to be undertaken can be broken down like this:

1 - Create model for storing Rules
2 - Create views to return rules, either entire articles or specific sections from an article
3 - Migrate enough data from API V1 to adaquetly test this new endpoint

calumbell:
Closed with #560

PR #560, opened 2024-10-08
Feature: V2 `/rules` endpoint

calumbell:
This PR goes part way to resolving issue #544 by introducing new models and views to API V2 for returning data equivelent to the `/sections` endpoint from V1 of the API.

## Wait, why `RuleGroup`?

Following discussion on the Open5e Discord, it was decided that the names of these endpoints should be `/rules` for specific rules and `/rulesets` for set of rules. However, the endpoint `/rulesets` is currently being used to return a list of different 'flavours' of 5e. D&D 2014, Level Up Advanced 5e, that kind of thing. It was decided that this endpoint should be renamed to `/gamesystem`, so that we can use `/ruleset` for sets of rules.

So if that was decided, how come I used the endpoint `/rulegroup`? The reason is that changing the endpoint for `/rulesets` -> `/gamesystems` would require not only updating models, views, serializers, &c., but also updating a lot of the V2 data, and also updating the V1->V2 converters. In other words, it would be a monster of a PR touching many different parts of the code-base, and likely be an absolute nightmare to review.

So `/rulegroup` is a temporary name used to split this issue into managable chunks


augustjohnson:
I have refactored the old ruleset to gamesystem.  Pull in changes from staging and adjust your code to use the different ruleset wording.

augustjohnson (changes requested):
Sending this back for now, merge in staging to your dev branch and refactor rulegroup to ruleset.

You'll probably have to pipenv install -d because of the new dependencies.

calumbell:
@augustjohnson Thanks for making the change from `RuleSet` -> `GameSystem`. Requested changes to this PR have been made. `RuleGroup` is now `RuleSet` and works as intended. Merging from `staging` proved painless! Let me know if you'd like anything else done before approving!

api_v2/models/rules.py
augustjohnson, line 1: filename should be singular, rule.py.

api_v2/models/rules.py
```diff
@@ -0,0 +1,37 @@
+from django.db import models
+from .abstracts import HasName, HasDescription, key_field
+from .document import FromDocument
+
+class RuleSet(HasName, HasDescription, FromDocument):
+  """
+  The RuleGroup model contains a set of Rules as part of a larger article, or
```
augustjohnson, line 7: Still a reference to ruleGroup here. Edit comments.

augustjohnson (changes requested):
Just a couple of very small changes.
* models/rules.py should be renamed rule.py
* any (comment) references to RuleGroup should be rewritten. I saw 2.

augustjohnson (approved):

Issue #586, opened 2024-10-20
Reliable OpenAPI Spec Generated


augustjohnson:
We require the ability to make an openapi spec reliably and accurately for all endpoints within the open5e api.

### Requirements
A valid OpenAPI Spec can be generated fairly quickly and easily.
It can be in either JSON or Yaml format.
It needs to be reasonably accurate. (/search has some issues, so we can consider those separate).

### Considerations
I have been using: `python manage.py generateschema`  but this appears to be going towards deprecation. Probably not a good long term solution. https://www.django-rest-framework.org/api-guide/schemas/#schema

Some of the v1 endpoints and v2 endpoints have a conflicting operationId when using that command. I have attempted to resolve that by specifying a base with the AutoSchema class. https://github.com/open5e/open5e-api/blob/staging/api/views.py#L143

Assuming we can reliably generate an accurate schema from code, we may NOT want to check it in to version control, because it would quickly become out of date (without discipline that we simply don't have as an open source project).

I also generally don't like adding dependencies into the project, less is always better. If it's possible to complete this using drf-spectacular, but ONLY within dev-dependencies that would be great.

augustjohnson:
I think the requirements for this will be roughly:
1. Generate a good openapi spec as part of the build (if not already).
2. Replace the standard drf UI with redoc at the root.
3. don't rely on the CDNs for this stuff, it should be self contained.

Issue #564, opened 2024-10-09
Standardize approach to weight across all of v2, similar to distance.


augustjohnson:
There should be a document-specific default weight and that should be displayed by default next to any weight field.

augustjohnson:
@calumbell this may be the type of change we'd queue up before launch due to it perhaps resulting in a json change. I see this as a prereq for i18n, which I think we'd like v2 to support.

For a bit of clarification: on a spell we store "range" as a decimal. We then do a lookup to see what unit the parent document uses as it's distance measurement unit, and then compute a "range_text" field to be "30 feet" while also making "range" the decimal available.

As of this moment, all objects have a weight field that's described as "weight" but no "weight_text" field is specified, and there is no weight unit specified at the document level

Also, everything that inherits from "object" has a weight (this includes creature).

PR #589, opened 2024-10-21
OAS Generation for APIV2

Dashron:
Addresses https://github.com/open5e/open5e-api/issues/586.

Complete
- Switches to drf-spectacular for oas generation
- Works for APIV2

Todo
- Fix all warnings & errors in oas generation
- Trim down the oas file to only include certain paths
- Set up spectacular to be installed only when the --dev flag is set


augustjohnson:
Adding to todo:
Update the github action to use this new method.

Dashron:
@augustjohnson I just pushed an update that eliminates most warnings and all errors. It adds annotations to clarify some types.

I left todo's for any type that tripped up my IDE. There are many fields VS code interpreted as `any` or `| none`. I think it's more of an IDE issue though, e.g. `HasSpeed` clearly types `walk` as float in the definition, but as `any` inside `get_speed`. :shrug: . 

I don't know which of these can be improved via code, and which need to be overridden via annotations, but I need to at the very least test each one to make sure my typing is correct.

I'll work on the final warnings and todo's tonight, then QA tomorrow.

Dashron:
When I run this, I get...

```
[truncated]/lib/python3.11/site-packages/rest_framework/fields.py:990: UserWarning: max_value should be a Decimal instance.
  warnings.warn("max_value should be a Decimal instance.")
[truncated]/lib/python3.11/site-packages/rest_framework/fields.py:992: UserWarning: min_value should be a Decimal instance.
  warnings.warn("min_value should be a Decimal instance.")
```

The error kind of looks like a warning that there are `model.DecimalField`s without min and max values set? I'm going to ignore this for now, but can create a ticket if you would like.

Dashron:
I also get a warning for this line: https://github.com/open5e/open5e-api/blob/staging/api_v2/serializers/spell.py#L23. Every other field I've worked with had a similarly named function or model attribute. This seems to be the only use of `slot_expended` in the code. Am I overlooking something or is it a bug?

Dashron:
It doesn't like the way [damage die types are created](https://github.com/open5e/open5e-api/blob/staging/api_v2/models/creature.py#L257-L271). Because multiple identical enums are created, it gets confused in naming them. This warning is more about highlighting a "possible" bug, not an issue in the output schema, so it's likely safe to ignore.

Dashron:
Ok, I just squashed the final warning by adding a manual postprocessing step to fix the search object.

@augustjohnson let me know if you want me to try to tackle [the decimal min/max issue](https://github.com/open5e/open5e-api/pull/589#issuecomment-2427924258), or remove [slot_expended](https://github.com/open5e/open5e-api/pull/589#issuecomment-2427926826). With those two wrapped up I would be comfortable with marking this ready for review.

augustjohnson:
Doing a bit of research into the decimal issue, I believe it's a DRF serializer throwing the error, and it's not clear which. 

The following places are all places where a decimal (or float) field are defined with mins that are integers instead of decimal/floats.  Adding a .0 to these would be easy enough.
https://github.com/open5e/open5e-api/blob/staging/api_v2/models/creature.py#L72
https://github.com/open5e/open5e-api/blob/staging/api_v2/models/abstracts.py#L52
https://github.com/open5e/open5e-api/blob/staging/api_v2/models/object.py#L29.
I have a related ticket here: https://github.com/open5e/open5e-api/issues/564

You have my blessing to add those changes in to see if it fixes the error.

Regarding the Extra Damage thing: I have no strong attachment to the current method, but it's a fairly established concept.  See Deva as an example of a creature who's attack has "extra" damage.
https://open5e.com/monsters/deva.  Feel free to simply put your concerns into an issue in the backlog and push as-is.

Dashron:
- removed slot_expended
- Figured out decimal, I have to wrap the number in `decimal.Decimal(x.x)`
- the extra damage warning went away? :shrug: 
- Unfortunately I don't think I can make this dev because it depends on annotations in the code.
- ironing out the github action update now

augustjohnson:
Hey, here's a thought.  Can we replace the current HTML UI here: https://api-beta.open5e.com/ with something nice generated by spectacular? I understand that competes with the readme featureset, but it's easier to find at our own url.

augustjohnson (approved):

augustjohnson:
Wow, amazing work here, I didn't realize you were manually forcing the properties to be one of the field types.  Yikes.  Lots of work. I think this will be a fantastic approach moving forward.

Dashron:
Thanks! Looking at the OAS file, there's a lot to clean up. I'll poke away at a couple more prs to work on that.

Issue #577, opened 2024-10-19
Server has N+1 problem


tylereed:
When hitting an endpoint, the server is generating multiple SQL queries for each spell in the result set.  This is because data linked through a foreign key is being accessed without telling django that it will be used.

When hitting `https://api.open5e.com/v2/spells`, 203 queries are generated.
When hitting `https://api.open5e.com/v2/spells?depth=1`, 253 queries are generated.

This is probably an issue with more than just spells, and is probably happening for other endpoints.

This is the classic N+1 problem, where the server is generating multiple queries.

calumbell:
With our recent crop of PRs (pending review as  of 31 Mar 2025) I think we have managed to resolve all the N+1 problems on the server.

Once PRs #656, #657, #658, and #659 are merged we should be good to close this issue

calumbell:
Looks like the N+1 problem has been fixed on all endpoints!

PR #592, opened 2024-10-21
Fixing up spell endpoint to fetch all needed data beforehand

tylereed:
Added in the related tables that were being fetched row by row when querying spells.  Locally, the speed went from 263 ms to 51 ms with these changes.  Data returned is the same, but performs 5 queries versus >250.

We should be able to apply this to other endpoints, but I first wanted to test this change out to see if it's working better.


tylereed:
This affects #577

calumbell (approved):
PR runs well on local. Observing a noticable improvement in response times when pulling data `v2/spells`. Code looks good, although popping in some short comments exlaining the two new methods would be helpful

PR #603, opened 2024-11-11
Bugfix (API V2): Added missing Serializers to `CreatureSerializer`

calumbell:
This PR fixes a bug where the `environments` and `languages` fields of the `/creatures` endpoint could not be filtered via query parameter.

eg. `v2/creature/?environment__fields=name` and  `v2/creature/?languages__field=name` had no effect on the API response.

This was because the `LanguageSerializer` and `EnvironmentSerializer` were not included on the `CreatureSerializer`, so these fields did not correctly inherit the logic defined in the `GameContentSerializer` abstract class for managing filtering nested fields.

Now, the query parameters for a creature of the form `/?fields=languages,environments&languages__fields=name&environments__fields=name` will generate a response of the form:

```
{
    "languages": [
        {
            "name": "Common"
        },
        {
            "name": "Draconic"
        },
        ...
     ],
    "environments": [
        {
            "name": "Arctic or Tundra"
        },
        {
            "name": "Caves"
        },
        ...
    ],
```

### DIFF
--- a/api_v2/serializers/creature.py
+++ b/api_v2/serializers/creature.py
@@ -8,6 +8,8 @@ from api_v2 import models
 
 from .abstracts import GameContentSerializer
 from .document import DocumentSerializer
+from .language import LanguageSerializer
+from .environment import EnvironmentSerializer
 from .size import SizeSerializer
 from drf_spectacular.utils import extend_schema_field, inline_serializer
 from drf_spectacular.types import OpenApiTypes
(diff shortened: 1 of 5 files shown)


calumbell:
Ah looks like, this is failing tests. To be expected, as it does change what is returned by the API when no parameters are provided. Should be a simply a case of updating the tests to match the new schema

calumbell:
(Tested updated and passing, good to go!)

augustjohnson (approved):

PR #608, opened 2024-11-18
V2 Creature Traits

### LINKED ISSUE #541: API V2 - `/creatures` endpoint does not return Creature Traits
*Creature traits are not implemented across the board, so for this example I will be using the [Aboleth from the WOTC SRD](https://api.open5e.com/v2/creatures/srd_aboleth/), which I believe has been set to use as something like a 'test monster'*

For clarity, Creature Traits are the features listed between the sections for a creature's proficiencies/senses/CR and its actions in a typical 5e statblock. Using the [the Aboleth's page on DND Beyond](https://www.dndbeyond.com/monsters/16762-aboleth) for reference, its creature traits are; Amphibious, Mucous Cloud and Probing Telepathy.

These traits already exist in the [V2 data](https://github.com/open5e/open5e-api/blob/staging/data/v2/wizards-of-the-coast/srd/CreatureTrait.json), there is also already a CreatureTrait model defined [in api_v2/models/creatures](https://github.com/open5e/open5e-api/blob/ed22a65609a3c25fb03dd4db6997ccdd9e9ecd5d/api_v2/models/creature.py) (starts line 245). So getting this feature up and running is probably just a case of configuring the creature serializer to return this data.

calumbell:
This PR closes #541 by updating the `v2/creatures` endpoint to also return a creature's traits. As this information was already part of the dataset it was simply a case of creating a new Serializer and ViewSet based on the `CreatureTrait` model.

Additional code was added to the `/creatures` view to avoid an N+1 error on th new `traits` fields, see issue #577


calumbell:
Ah, of course this would explode the tests. It changes the shape of the API response for the `/creatures` endpoint by adding the `traits` field so that it no longer matches the expected result. Will go ahead and update that

augustjohnson (approved):

PR #614, opened 2024-12-20
610 add extra column to class table on character classes

augustjohnson:


augustjohnson:
There's a bug with this qs:http://localhost:8000/v2/spells/?limit=50&page=1&document__key__in=open5e-e,open5e,srd&ordering=name&fields=key,name,document,level,school,classes&document__fields=name,key&classes__fields=name&school__fields=name,key&depth=1&name__contains=&school__key=&classes__key__in=

PR #620, opened 2025-02-17
Update Monster.json

postrowski:
more cases of "\n" that should have been "\\n"


eepMoody:
Ah looks like you might have hit some of the same lines as @calumbell, any chance you can clean that conflict up, and then I can get this merged in!

Thank you for the fixes, much appreciated!

Issue #618, opened 2025-02-15
API V2: `damage_resistances`, `damage_immunities` and `condition_immunities` fields of `/creature` endpoint don't return useful information


calumbell:
# Bug Description

(The TLDR)

**Observed Behaviour** The `v2/creatures` endpoint doesn't return information about a creatures damage_immunities, damage_resistances or condition_immunities. It only returns the API URL for these pages.

**Expected Behaviour**The `v2/creatures` endpoint should be able to display, at a minimum, the name of the damage type (in plain text) and the description of its damage resistances, damages immunities and condition immunities.

## Recreation 

For this example I will the Aboleth, Nihilith statblock from Tome of Beasts as damage resistances and damange/condition immunities.

First, fire up the `staging` branch of the Open5e API on your local server and navigate to the Aboleth, Nihilith V2 page (with params for filtering only relevant information):

`/v2/creatures/tob_aboleth-nihilith/?fields=name,damage_resistances`
```
{
    "name": "Aboleth, Nihilith",
    "damage_resistances": [
        "http://localhost:8000/v2/damagetypes/acid/",
        "http://localhost:8000/v2/damagetypes/bludgeoning/",
        "http://localhost:8000/v2/damagetypes/fire/",
        "http://localhost:8000/v2/damagetypes/lightning/",
        "http://localhost:8000/v2/damagetypes/piercing/",
        "http://localhost:8000/v2/damagetypes/slashing/"
    ]
}
```

The `damage_resistances` fields returns an array of API URLs for the different damage types. This is not very helpful. There doesn't be a way to coerce the API into making these nested queries via the `creatures` endpoint, in the same way that you can do so with the `document` field via query parameters:

`/v2/creatures/tob_aboleth-nihilith/?fields=name,document`  (no filtering, similar to above example)

```
{
    "document": "http://localhost:8000/v2/documents/tob/",
    "name": "Aboleth, Nihilith"
}
```

### Depth parameter
The `document` field is joined when the `creatures` endpoint is accessed with a depth parameter is 1 or greater

`/v2/creatures/tob_aboleth-nihilith/?fields=name,document&depth=1`
```
{
    "url": "http://localhost:8000/v2/documents/tob/",
    "key": "tob",
    "licenses": [ ... ],
    "publisher": "http://localhost:8000/v2/publishers/kobold-press/",
    "gamesystem": "http://localhost:8000/v2/gamesystems/o5e/",
     "name": "Tome of Beasts",

}
```

[...]

PR #622, opened 2025-02-21
Bugfix: improve results for nested fields returned by `/v2/creatures`

calumbell:
This feautre resolves issue #618 by adding missing nested serializers to the V2 `CreatureSerializer` for the following fields:

- `conditions_immunities`
- `damage_immunities`
- `damage_resistances`
- `damage_vulnerabilities`

This inclusion means that these fields on the `creatures` endpoint will return more helpful fields than just their URL.

For example, on the staging `branch` the query `/v2/creatures/bfrd_aboleth/?fields=name,damage_resistances` currently returns:

```
{
  "name": "Aboleth",
  "damage_resistances": [
         "http://localhost:8000/v2/damagetypes/acid/"
   ]
}
```

The URL returned by the `damage_resistances` field is not especially helpful.

This PR changes the result this same query returns to:
```
{
    "name": "Aboleth",
    "damage_resistances": [
        {
            "url": "http://localhost:8000/v2/damagetypes/acid/",
            "key": "acid",
            "name": "Acid",
            "desc": "The corrosive spray of a black dragon’s breath and the dissolving enzymes secreted by a black pudding deal acid damage.",
            "document": "http://localhost:8000/v2/documents/srd/"
        }
    ]
}
```

Much better. If this is too verbose then the fields returned by the `damage_resistances` field can be modified using the `damage_resistances__fields` parameter.

`/v2/creatures/bfrd_aboleth/?fields=name,damage_resistances&damage_resistances__fields=name`
```
{
    "name": "Aboleth",
    "damage_resistances": [
        {
            "name": "Acid"
        }
    ]
}
```

The `damage_immunities`, `damage_vulnerabilities` and `condition_immunities` fields function the same way.

It looks like these fields are already being pre-fetched by the Creatures view, which should nip any N+1 errors in the bud

### DIFF
--- a/api_v2/serializers/creature.py
+++ b/api_v2/serializers/creature.py
@@ -67,6 +67,7 @@ class CreatureSerializer(GameContentSerializer):
     skill_bonuses = serializers.SerializerMethodField()
     skill_bonuses_all = serializers.SerializerMethodField()
     damage_immunities = DamageTypeSerializer(many=True)
+    damage_resistances = DamageTypeSerializer(many=True)
     actions = CreatureActionSerializer(many=True, context={'request': {}})
     traits = CreatureTraitSerializer(many=True, read_only=True)
     speed = serializers.SerializerMethodField()
--- a/api_v2/serializers/creature.py
+++ b/api_v2/serializers/creature.py
@@ -8,6 +8,7 @@ from api_v2 import models
 
 from .abstracts import GameContentSerializer
 from .damagetype import DamageTypeSerializer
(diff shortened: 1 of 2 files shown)


calumbell:
Closes #618

calumbell:
OK, I have updated the test that runs on the `/v2/creatures` route. But it looks like we have some test failing. Checking out the `staging` branch, those same tests are still failing. Not sure how that one slipped through the net, i'll open a new issue

augustjohnson (approved):

PR #643, opened 2025-03-06
Replace `DocumentSerializer` with `DocumentSummarySerializer` in some contexts

### LINKED ISSUE #637: Most endpoints return the entire documents object.
document details is probably not required when viewing an individual spell.  @calumbell  and I have discussed a solution to this and #636 that is a new convention of "DocumentSummarySerializer" being a new thing that returns a subset of fields.

calumbell:
Closes #637 

This PR adds the new `DocumentSummarySerializer` (created as part of PR #641) to many other serializers which only which previously used the `DocumentSerializer`, but really didn't need to have access to absolutely ALL the information returned by the Documents endpoint.

Additional work was undertaken to set up eager loading on views that were effected by these new serializers. This was to resolve any N+1 problems before they occur (as per issue #577).

The response times and the number of SQL queries required to generate those responses was reduced noticably on the updated endpoints

## Next Steps

- There is lots of duplicate code for setting up eager loading. It would be a good idea to find a way to componentise this logic to keep things DRY and maintainable.
- Some endpoints might benefit from an even smaller Document serializer, perhaps with only the `name` and `key` fields. I am looking specicially at the `'languages'` field returned by the `'v2/creatures'` endpoint. ie.

From `/v2/creautres/srd_goblin`
```
 "languages": [
        {
            "url": "http://localhost:8000/v2/languages/common/",
            "key": "common",
            "document": {
                "name": "Systems Reference Document",
                "key": "srd",
                "publisher": {
                    "name": "Wizards of the Coast",
                    "key": "wizards-of-the-coast"
                },
                "gamesystem": {
                    "name": "5th Edition",
                    "key": "o5e"
                },
                "permalink": "https://dnd.wizards.com/resources/systems-reference-document"
            },
            "name": "Common",
            "desc": "Typical speakers are Humans.",
            "is_exotic": false,
            "is_secret": false,
            "script_language": "http://localhost:8000/v2/languages/common/"
        },
        {
            "url": "http://localhost:8000/v2/languages/goblin/",
            "key": "goblin",
            "document": {
                "name": "Systems Reference Document",
                "key": "srd",
                "publisher": {
                    "name": "Wizards of the Coast",
                    "key": "wizards-of-the-coast"
                },
                "gamesystem": {
                    "name": "5th Edition",
                    "key": "o5e"
                },
                "permalink": "https://dnd.wizards.com/resources/systems-reference-document"
            },
            "name": "Goblin",
            "desc": "Typical speakers are goblinoids.",
            "is_exotic": false,
            "is_secret": false,
            "script_language": "http://localhost:8000/v2/languages/dwarvish/"
        }
    ],
```

This still feels kind of over the top to me


augustjohnson (approved):
Big step in the right direction.

Issue #645, opened 2025-03-07
Refactor eager loading code into `EagerLoadingMixin`


calumbell:
Currently, there is a lot of repetition in the eager loading code on our V2 Views. Refactoring this functionality in a reusable way would greatly improve maintainability of this feature.

Initial research suggests that a common way to handle this would be to implement this logic as a mixin that our View classes will inherit from in addition to the `ReadOnlyModelViewSet`

# Work Required
- Create a `EagerLoadingMixin` that recreates current Eager Loading functionality
- Refactor V2 Views use the `EagerLoadMixin` to handle their eager loading.

PR #647, opened 2025-03-07
Refactor eager loading code into `EagerLoadingMixin`

calumbell:
This PR goes half way to fixing #645 by creating an `EagerLoadingMixin` that encapsulates eager loading logic across viewsets.

Dynamically applies `selected_related()` for ForeignKey fields and `prefetch_related()` from ManyToMany/reverse relationships. This improves query efficiency and prevents N+1 problems
  
  ## Usage
  1. Make sure your ViewSet inherits from `EagerLoadingMixin` before its base
  class (ie. ReadOnlyModelViewSet).
  2. Re-define `select_related_fields` and `prefetch_related_fields` lists on 
  the child ViewSet to specify relationships to optimise.
  
  ## Example
  ```
    # EagerLoadingMixin inhertired before base-case

    class CreatureViewSet(EagerLoadingMixin, viewsets.ReadOnlyModelViewSet):
      queryset = models.Creature.objects.all().order_by('pk')
      serializer_class = serializers.CreatureSerializer
      filterset_class = CreatureFilterSet

      # ForeignKey relations to optimise with select_related()
      select_related_fields = []

      # ManyToMany / reverse relations to optimise with prefetch_related()
      prefetch_related_fields = []
  ```

### DIFF
--- a/api_v2/views/mixins.py
+++ b/api_v2/views/mixins.py
@@ -1,0 +1,59 @@
+class EagerLoadingMixin:
+  """
+  Mixin to apply eager loading optimisations to a ViewSet.
+  
+  Dynamically applies `selected_related()` for ForeignKey fields and 
+  `prefetch_related()` from ManyToMany/reverse relationships. This improves 
+  query efficiency and prevents N+1 problems
+  
+  ## Usage
+  1. Make sure your ViewSet inherits from `EagerLoadingMixin` before its base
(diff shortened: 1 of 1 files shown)


augustjohnson:
This looks good, we need to ensure it doesn't interfere with search, which has a totally distinct parameter setup.

augustjohnson (approved):

PR #648, opened 2025-03-07
EepMoody/issue644

### LINKED ISSUE #644: Add `"initiative_bonus"` field to Creature model/data
Following on from discussion on the Open5e Discord w.r.t. looser way in which the Monster Manual 2024 handles monster initiative bonuses, I feel like there is a case for returning this data from our API seperately from the dexterity mod

On the Creature model we would like to add an `"initiative_bonus"` field. This field should be populated with whatever value is found in the similarly named and similarly not-yet-existant `"initiative_bonus"` in the Creature V2 data.

If not initiative information is included in a creature's source JSON, the `"initiative_bonus"` field on the Creature model should be set to the Creature's dexterity modifier

eepMoody:
Adds initiative_bonus field. 

If this field is included in the JSON the provided value will be used, otherwise it will fall back to dexterity bonus.

closes #644


calumbell (approved):
Welcome back! Looks good, runs as intended on local. I'll go ahead and approve.

augustjohnson (approved):

PR #649, opened 2025-03-08
replace \r\n with \n\n globally

### LINKED ISSUE #646: V2 Data: SRD Class starting proficiencies not formatted correctly.
# Bug Description

The markdown description for starting class proficiencies for all SRD base-classes doesn't insert new lines where there ought to be new lines

## Bug Recreation

1. Fork the current `staging` branch of the Open5e website. Start up the test server. Make sure it is pulling data from the test Django server also running the `staging` branch.
2. Visit any of the SRD Class pages via the `/classes` page.
3. Notice that the **Proficiences** section looks off. It is missing line-breaks:

Image

## Possible Solution

The problem likely lies in the `data/v2/wizards-of-the-coast/srd/ClassFeature.json` file. This is where to markdown used for starting proficiencies is defined.

Looking at the Barbarian:
```
{
  "model": "api_v2.classfeature",
  "pk": "srd_barbarian_proficiencies",
  "fields": {
    "name": "Proficiencies",
    "desc": "**Armor:** Light armor, medium armor, shields\r\n**Weapons:** Simple weapons, martial weapons\r\n**Tools:** None\r\n**Saving Throws:** Strength, Constitution\r\n**Skills:** Choose two from Animal Handling, Athletics, Intimidation, Nature, Perception, and Survival",
    "document": "srd",
    "parent": "srd_barbarian"
  }
},
```

Doesn't create a line break `"\r\n"`. It should be `"\n\n"`. A similar problem exists on the other SRD Classes

eepMoody:
Closes #646

Replaces `\r\n` with `\n\n` to correctly render newlines


augustjohnson (approved):

augustjohnson:
I prefer not to add new libraries whenever possible. It increases complexity. Approving for now, but we should consider refactoring to reduce reliance on tabulate.

augustjohnson (approved):

eepMoody:
After some discussion, I think maybe we should lean on the UI fixes instead for now, and take a more surgical approach to cleaning up any data issues. I'm gonna go ahead and close this and we can revisit later.

That said, @augustjohnson  I do think that something like Tabulate would make sense for doing general data cleanup. Markdown tables are finnicky. I was going in and adding cases to cover all the various issues in our existing tables, but it was quickly ballooning into a mess of regex and special cases.

I think the ideal would be some small set of highly-targeted but robust formatting libraries, and run them on a pre-commit hook or something? Make sure that any data going into the system is consistently formatted.

PR #656, opened 2025-03-25
Fixed N+1 problem on `/v2/classes`

calumbell:
This PR goes some way to addressing issue #577 by fixing N+1 errors on the 

Here is a before and after for hitting the `/v2/classes` API endpoint.

- **`Staging`**. Query takes ~2800 ms to execute, 605 SQL queries are made
- **Branch**. Query takes ~800 ms to execute, 22 SQL queries are made

So a substantial improvement!

Previously, when implement prefetching/eager loading of related resources, the fetching of ClassFeatureItems could not be optimised because the fields that returned ClassFeatureItems on the API; `"gained_at"` and `"column_data"`, were properties defined on the model. These props did not exist in the DB, so could not be prefetched.

My solution to move the definition of the derived `"gained_at"` and `"column_data"` fields from the model to the `to_representation()` method on the serializer, and then creating a seperate `ClassFeaturePrefetchSerializer` to serialize the `"feature_items"` field on the ClassFeature model. This data can now be prefetched using the `EagerLoadingMixin`


calumbell:
(Closing and re-opening to re-run tests)

augustjohnson (approved):

PR #657, opened 2025-03-26
Fixed N+1 problem on `/v2/creatures` endpoint

calumbell:
w.r.t. #577 

This PR refactors the models, serializers, and views associated with the `/v2/creatures` endpoint to better support field prefetching. This is used to mitigate a pronounced N+1 problem on this endpoint that was leading to sluggish queries.

**BENCHMARK**

*Before*. The `/v2/creatures` API endpoint would take ~3500 ms to respond, and made 463 SQL queries to the DB

*After*. The `/v2/creatures` API endpoint takes ~950 ms to responde, and makes 27 SQL queries to the DB.

Ideally, we would like to explore further ways to improve these response times. But this is a set in the right direction!


calumbell:
Closing and re-opening to re-run validation

augustjohnson (approved):

PR #662, opened 2025-03-31
Added `"armor_detail"` to Creature model

### LINKED ISSUE #600: Natural Armor needs a flag.
Creatures sometimes have "natural armor" as a tag next to their armor class.  This should be a flag in the API.
This work includes added the field and the data.

calumbell:
## Description
This PR adds the `"armor_detail"` field to the V2 Creature model, replicating the functionality of the "`armor_desc`" field of the V1 Monster model.

This field is designed to hold the data in a monster's statblock that sometimes follows their armor class in paratheticals ie. the *(natural armor)* tag in the WotC 5.1 SRD Aboleth statblock.

425739397-78fca35a-3e7e-48e8-bbe2-109bd49e6f6c

Additional work was undertaken to convert all existing v1 `"armor_desc"` data to v2 `"armor_detail"`. A simple Python script was written to achieve this. I have included this as part of my PR so we have a record of it, but is not run as part of any automated build/setup step, was only designed to be executed once from the command line, and can be deleted should we choose.

## Type of Change

Enchancement (adding back missing V1 feature)

## Related Issue

Closes #600

## How has this been tested?

The fix was first tested by spinning up a local Django test server and poking around the `/v2/creatures` endpoint. Monsters from many different source documents were examined this way.

Because this PR changes the shape of the API response of anything with an armor class, running `pytest` flagged several tests as failing. Once the test cases were updated to better reflect the new shape of the API response, all tests were passing.

## Notes

Once this PR has been merged to `staging`, we will want to update the front-end to display this additional information on the `/monsters/[id]` route


calumbell:
Looks like a weird character found its way into the data. Passed tests on local, but failed on PR validation. Fixed the problem that was flagged, but there might be more. Closing and re-opening PR to re-run validation

calumbell:
Ah looks like this is a data validation issue, not a testing one. Had a go sanitising unicode characters from API V2 data. Tests are passing locally, so hoping this clears validation. Closing and re-opening to re-run

augustjohnson (approved):

PR #663, opened 2025-04-05
653 icons for v2 objects

### LINKED ISSUE #653: Icons for v2 objects
Consider adding basic icon functionality for v2 objects.
Sources that appear to be appropriately licensed:

https://www.elderberry-inn.com/icons
https://game-icons.net/

Design wise, I think images would be a distinct application, and each object could have a lookup that resolves an image for it based on some traits.

augustjohnson:
Adding in icons on Conditions.


calumbell (changes requested):
Hey, so this PR and the issue it resolves are both a little light on detail. Is the plan to host all Open5e images in the static folder, or is this a temporary solution? It might make more sense to keep this with our other data.

One problem that I noticed is that this PR introduces an N+1 problem on the `/conditions` endpoint:

Screenshot 2025-04-09 at 11 34 07

With 15 conditions, the images table is being queried 15 times. This can be streamlined using `EagerLoadingMixin` on the conditions view and configuring it to eagerly load the image FKs

augustjohnson:
I'll adjust and use eagerloadingmixin.

As far as the feature itself is concerned, I intended to create a new data type, "Image". This is metadata describing something that can be presented as an image on the web, url, document credit, license, etc.

I also found a few high quality sources of icons that are appropriately licensed, but the first i was going to use was the elderberry inn icon set (specifically the conditions). They are monocolor SVGs and could almost certainly get integrated into the main UI one way or another, or into any integrators.

There's also the illustrations created for the v1 api: https://open5e.com/monsters/aboleth. I was going to extend the usage of the "image" object to also include an optional "illustration" field that can be added to any game content object. I feel like this is currently a missing feature, and I built the Image object to be able to fulfill this use case as well.

There's a complex process of linking an individual game content object to an icon, but I think various scripts can help do that in the offline dev process, for example matching "goblin.svg" to all different "goblin" creatures shouldn't be TOO hard.

calumbell (approved):
Looks good! Not seeing those N+1 errors anymore.

Curious about how this will interact with our existing source selection system on the front-end, but that is a front-end problem.

Getting images on the API is exciting!

PR #672, opened 2025-04-16
update url for docker hub builds so it is visible

eepMoody:
This PR updates the URL for DockerHub deploys to point to an actual functioning URL.

That's all.


eepMoody:
Turns out this isn't really a fixable issue, so will just leave it :(

PR #679, opened 2025-04-18
Adding a dirty check, and a --clean command.

### LINKED ISSUE #527: Quicksetup should include a check if the dir is dirty
The quicksetup command could be refreshed to include a few things:
1. If a whoosh_index folder exists
2. If a static_files folder exists
3. If a db.sqlite file exists
4. If there are any model/migration mismatches

Warn or maybe delete depending on the flags.

augustjohnson:


augustjohnson:
This adds a check for dirty directories in quicksetup (occasionally a problem if you're switching branches locally), and a flag to clean up 3 dirty spots. Tested on my directories locally (debian). I'd be interested if a mac or windows user could test it, but it's unlikely to be used by very many.

PR #683, opened 2025-04-18
Moved logo.png into static, and fixed readme.md reference.

### LINKED ISSUE #673: Move logo.png into static
logo.png should not be in the root, it should be in static. Move it, figure out what that breaks, then fix all of it.

augustjohnson:
I prefer a cleaner root directory, and this puts the logo into the static dir.
I searched for all other references to logo.png, readme was the only one.
Yes, this will probably deploy the 25.3kb file onto our db server. I am unconcerned.


calumbell (approved):
Looks good, doesn't seem to break anything. It might seem like not a big deal, but I think there is value in keeping the root directory as lean as possible 

Although we aren't using the logo from the BE on the FE or anything, there are definetly contexts that I can imagine where having the logo available on the server might be helpful.

PR #698, opened 2025-04-22
[BUGFIX] Languages field on Creatures endpoint now work as intended

### LINKED ISSUE #669: V2: Creature language field listing returning all languages in DB
## Description

There is currently an issue with the `"languages"` field of the `/v2/creatures/` endpoint on the `staging` branch of the Open5e API.

When the languages a monster knows are listed in a source statblock read something like `"any one language (usually Common)"` or `"any four languages"`, the API representation of that same statblock will return a list of every single language in the Open5e dataset for its `'language'` field

If we take a look at the Veteran statblock from the WotC 5.1 SRD (pp. 403), we can see that that the **Languages** field has the value "any one language (usually Common):

Image

However, if we look at the page on the Open5e website (the `staging` branch, pulling data from a local Django server running the `staging` branch) we can see that the Languages field is populated with what looks like every language in the Open5e dataset:

Image

Quite the polyglot.

I see the intention with this design choice: the veteran knows **any** one language, so they *could* speak any of the languages in the O5e DB. But the way it is presented in the data is not only inaccurate (it looks like they can speak them ALL) and kind of goofy (really, really long), but it really bloats the size of the JSON response (take a look at `v2/creatures/srd_veteran/`, roughly half of the JSON payload is languages),  which will slow down response times and increase server resource usage.

## Proposed Solution

Currently, the `languages` field of the `Creature` model is a ManyToMany relationship to the `Language` model. So as it stands there is currently no way to store any supplementry plain-text information about the languages a creature speaks.

One idea (far from the only one!) might be to create a field on the `Creature` model with a name like `"languages_template"` that will store a string to display. If a language is mentioned in the `"language_template"`, perhaps we can use some kind of template syntax to insert data from the `"languages"` list.

```
"languages_template": "any one language (usually {{ common }})",
"languages": [
  {
    "name": "Common",
    "key": "common",
    "desc": "Typical speakers are Humans.",
  }
]
```

If `"language_template"` is null(-ish), the list of languages can be rendered as is.

Another idea would be to lose the templating, but run a little code on the client that scans the `"language_template"` string to languages that occur in the `"languages"` array and then insert any additional contextual information (ie. showing the contents of the `"desc"` on a hover)

calumbell:
## Description
This PR address two bugs present on the `/v2/creatures` endpoint on the `"languages"` field:

1) Plain-text data for how to render the Languages infomation on a creature's statblock wasn't being returned by the API.
2) If the text describing the creature's languages contained the words "all" or "any", then ALL languages from the `/v2/languages` endpoint would be returned.

Point one presents makes it difficult to reconstruct statblocks on the front-end, point two results in a massive overfetching of irrelevant data.

Point one was resolved by updating the serializers for the Creatures and Languages endpoints so that the Creature view returned data in the following structure:

```
    "languages": {
        "as_string": "any one language (usually Common)",
        "data": [
            {
                "name": "Common",
                "key": "common",
                "url": "http://localhost:8000/v2/languages/common/",
                "desc": "Typical speakers are Humans."
            }
        ]
    },
```

Point two was addressed by massaging the data, and removing a lot of erroneous languages included in our creature datasets.

Finally, tests were updated to reflect the new shape of the JSON response.

## Related Issue
Closes #669 

## Notes

This PR will likely cause breaking changes on the front-end `/monsters/[id]` page. The JS used to render the languages list will almost certainly need to be updated.


augustjohnson (approved):

augustjohnson:
@calumbell We got some merge conflicts after the other changes, could you sort them out when you get a chance?

calumbell:
@augustjohnson Sure! Looks like the conflicts are only in the tests so should be pretty straightforward to resolve

calumbell:
Fixed. Thanks for flagging

PR #716, opened 2025-04-27
V2: Rename `/races` endpoint to `/species`

### LINKED ISSUE #704: Rename "v2/races" to "v2/species"
In the new 2024 edition the problematic term "race" has been replaced by "species".

Our endpoint for v2 should reflect this adjustment.
v1 should (unfortunately) stay as-is to avoid breaking changes for existing users, but we will use the better term going forward.

calumbell:
## Description

This PR changes the name of the `/v2/races` endpoint to `/v2/species`. Attendent model, serializers, etc. names have also been updated to reflect this change.

## Related Issue

Closes #704


augustjohnson:
I think you will need to adjust a few lines in /search. This one for example:

https://github.com/open5e/open5e-api/blob/staging/search/serializers.py#L57

augustjohnson (changes requested):
https://github.com/open5e/open5e-api/blob/staging/search/serializers.py#L57
You will have to change a few lines in search to maintain functionality. Maybe just this one?

calumbell:
Good spot!! Will take a look

calumbell:
Right, looks like the `/search` endpoint is indexing items on the `/v2/species`. Let me know if you notice anything else weird!

augustjohnson (approved):

Issue #655, opened 2025-03-15
V2 Enhancement: additional contextual information for Creature resistances/vulnerabilities/immunities


calumbell:
# The Problem

Currently there is no way to include additional contextual information about a specific monster damage resistances/immunities - ie. whether they are limited by monster form, etc.

I encountered an issues while working on the resistances, vulnerabilities, etc., section of the single-view monster page on the website (as part of this issue https://github.com/open5e/open5e/issues/658). Certain monster statblocks - in this case `/v2/creatures/tob_aboleth-nihilith` had both 

Take a look at the following query:

`/v2/creatures/tob_aboleth-nihilith/?fields=name,nonmagical_attack_immunity,nonmagical_attack_resistance,damage_resistances,damage_immunities&damage_resistances__fields=name&damage_immunities__fields=name` (query params included to keep response brief).

```
{
    "name": "Aboleth, Nihilith",
    "damage_immunities": [
        {
            "name": "Bludgeoning"
        },
        {
            "name": "Cold"
        },
        {
            "name": "Necrotic"
        },
        {
            "name": "Piercing"
        },
        {
            "name": "Poison"
        },
        {
            "name": "Slashing"
        }
    ],
    "nonmagical_attack_immunity": true,
    "damage_resistances": [
        {
            "name": "Acid"
        },
        {
            "name": "Bludgeoning"
        },
        {
            "name": "Fire"
        },
        {
            "name": "Lightning"
        },
        {
            "name": "Piercing"
        },
        {
            "name": "Slashing"
        }
    ],
    "nonmagical_attack_resistance": true
}
```

This struck me as strange while working with this page query specifically – a monster being resistant to non-magical attacks is inconsequential if they are already immune to non-magical attacks, why bother including it in our data? Before ploughing on ahead, I took a peak at the [same statblock on the Kobold Press website](https://koboldpress.com/creature-codex-nihileth-aboleth/):

<img width="952" alt="Image" src="https://github.com/user-attachments/assets/b9c2f914-cc0c-4165-9d7c-43cc0b606a3e" />

If you look at the resistance and immunities, there is additional contextual information in parentheticals after certain resistances/immunities that go a long way to explain what is going on. This additional information is required to correctly render the statblock on the frontend, but the API doesn't currently return it.

## Propose Solution

There is some design work that will need to take place to figure out the best way of including, but the additional information will have to be stored on the `Creature` model.

eepMoody:
[...]

PR #717, opened 2025-04-28
Implement QA fixes on `/v2/creatures` endpoint

### LINKED ISSUE #707: Creature endpoint QA fixes
Based on #687 

- [ ] There seems to be some overlap between the `"nonmagical_attack_resistance/immunity"` fields and just adding `"slashing"`, `"bludgeoning"`, and `"piercing"` damage to the `"damage_resistances/vulnerabilities/immunities"` arrays.
- [ ] The `"parent"` fields returned by objects in the `"traits"` object is unneeded.
- [ ] CreatureTraits `"type"` field appears to be `null` for all entries. 
  - [ ] Verify that this is not a bug
  - Might be worth omitting the `traits.type` field from the View if it isn't returning useful information
- [ ] `"action.order"` field might benefit from a more verbose name to makes its intended use more clear.
  - eg. "`action.order_statblock`"
- [ ] `"action.form_condition"` field might benefit from a more verbose name to make its intended use more clear. 
   - eg. `"limited_to_form"`, but probably a bit better than that.
- [ ] [MINOR] changing `"legendary_cost"` to `"legendary_action_cost"` would make this field's function even more clear.
- [ ] The `"parent"` field  in the objects returned by the `"actions"` is not necessary in this View (we are already looking at the parent)
- [ ] The `"parent"` in the objects returned by the `"attacks"` field of the creature's `"actions"` is not necessary (the information is already there)
- [ ] There might be better names and JSON structure for the`"uses_type"` and "`uses_param`" so that it is more obvious what their indented function is. As best as I understand it, these fields place  For example:

Current:
```
[
	{
    	"uses_type": "PER_DAY",
    	"uses_param": 1,
	},
	{
    	"uses_type": "RECHARGE",
    	"uses_param": 6,    
	}
]
```

Candidate:
```
[
	{
  	"usage_limits": {
      	"type": "PER_DAY",
      	"param": 1,
  	}
	},
	{
    	"usage_limits": {
        	"type": "RECHARGE",
        	"param": 6,
    	}
	}
]
```

calumbell:
## Description

This PR fixes a number of small issues on the `/v2/creatures` endpoint that were recently highlighted on a walkthrough of the endpoint:

- Removed the unneeded `"parent"` field from the CreatureTraitsSerializer
- Removed the unneeded `"parent"` field from the CreatureActionSerializer
- Removed the unneeded `"parent"` field from the CreatureAttackActionSerializer
- Remove the unused `"type"` field from the CreatureTraits serializer
- `"action.order"` renamed to `"action.order_in_statblock"` to increase readibility of JSON payload
- `"action.form_condition"` renamed to `"action.limited_to_form"` to increase readibiltiy of JSON payload
- `"action.legendary_cost"` renamed to `"action.legendary_action_cost"` to increase readibilty of JSON payload
- Renamed and restructured the `"uses_type"` and `"uses_param"` fields to increase readibiltiy of JSON payload

The only point listed in the initial issue that I haven't addressed here is the first, which touches on the overlap between various damage resistance and immunity fields. This is a more complex change, and as it already has its own issue (#655), I feel it would be better handled in its own PR.

With this PR I have made an effort to avoid modifying the source data and the Django models, as this could well interfer with the workflow of those who are currently working to ingest data from the 2024 SRD. So, where a change of a field name has required I have implemented this on the serializer instead of the model. It might be worth it as some point to update the model and source data to reflect the new field names to simplify future data entry.

## Following Up

- The `"type"` field on the CreatureTrait model appears to be completely unused. It is `null` in all of our V2 data. We should consider using it for something or removing it.

## Related Issue

Closes #707 

## How was this tested?

A/B testing against `staging`. Runnging `pytest` once test case were updated to refelct new API response JSON structurre.


augustjohnson (approved):
This is a good change overall. I agree with the comment about field names. I will create an issue associated with that, but it's lower priority and not blocking. I think it helps with the data input process though.

Edit: #718

calumbell:
Good idea, i'll throw any discrepencies I come across/create onto the issue as comments

PR #720, opened 2025-04-29
564 standardize approach to weight across all of v2 similar to distance

### LINKED ISSUE #564: Standardize approach to weight across all of v2, similar to distance.

augustjohnson:
The intent of this PR is to provide a "weight_unit" field anywhere a weight is displayed. For right now, it's only settable once at the document level, and not overrideable like distance is. That can be an enhancement at some point in the future if necessary.

I added the field at the document level, exposed it in the relevant serializer (just item, creatures we don't make an assertion about weight), and checked the responses.


calumbell (approved):
Nice work. Definitely makes sense to refactor `"weight_unit"` to use an enum.

The Document model seems like a good place for this field to live, as I think it is safe to assume that distance units will be consistant across the same source. My instinct is that GameSystem is the best place to store this information, but this makes a lot of assumptions about how consistant units are across different sources under the same GameSystem. So it might be a horrible idea.

Tests are passing, everything looks good on the test server ✅

PR #721, opened 2025-04-30
Bugfix: Query Parameter Filtering

### LINKED ISSUE #719: [BUG] Filtering via Query Parameter is inconsistant
## Description

Filtering via query parameter, ie. adding a `/?fields=name,key,url` or `/?document__fields=name` works perfectly well in most cases, but not all.

I believe that the root of this problem is that not all Serializers on the Open5e API inherit from the abstract `GameContentSerializer`, which is where the logic that handles the filtering by query parameter is handled.

However, just updating the `serializers.ModelSerializer` to `GameContentSerializer` causes some fields to not serializer at all, and instead return their URL. 

After some initial digging and poking around at the code, I believe that the way that depth is handled in the `GameContentSerializer` is causing this issue

## Recreation

1) Spin up a local Django test server running an up-to-date version of the `staging` branch
2) Visit the `/v2/spells/srd_fireball/?fields=name,classes`, you will receive the following JSON: 

```
{
    "classes": [
        {
            "name": "Sorcerer",
            "key": "srd_sorcerer",
            "url": "http://localhost:8000/v2/classes/srd_sorcerer/"
        },
        {
            "name": "Wizard",
            "key": "srd_wizard",
            "url": "http://localhost:8000/v2/classes/srd_wizard/"
        }
    ],
    "name": "Fireball"
}
```

3) Now try to filter the the fields of the `"classes"` objects by updating the query parameters: `/v2/spells/srd_fireball/?fields=name,classes&classes__fields=name`. The addition of this additional query parameter has no effect on the returned JSON
4) Now compare this to the `"document"` field returned by the `/v2/spells` endpoint. Visit `/v2/spells/srd_fireball/?fields=name,document`

```
{
    "document": {
        "name": "Systems Reference Document",
        "key": "srd",
        "publisher": {
            "name": "Wizards of the Coast",
            "key": "wizards-of-the-coast",
            "url": "http://localhost:8000/v2/publishers/wizards-of-the-coast/
        },
        "gamesystem": {
            "name": "5th Edition",
            "key": "o5e",
            "url": "http://localhost:8000/v2/gamesystems/o5e/"
        },
        "permalink": "https://dnd.wizards.com/resources/systems-reference-document"
    },
    "name": "Fireball"
}
```

5) Now attempt to filter the fields returned by the `"document"` object via query parameters. ie. `/v2/spells/srd_fireball/?fields=name,document&document__fields=name` 

```
{
    "document": {
        "name": "Systems Reference Document"
    },
    "name": "Fireball"
}
```

## Observed Behaviour
Not all Serializer fields respect the `?fields` query parameter. Many do, but not all.

## Desired Behaviour
All Serializer fields should respect the `?fields` query parameter

calumbell:
## Description

This PR fixes and issue on the API where not all fields returned by an endpoint were filterable via query parameter.

At first I thought this would be a simple fix, just update all the Serailizers so that they all inherit from the `GameContentSerializer`, which is where our query parameter filtering is handled. There ended up being quite a few hurdles to jump than I anticipated, namely weird interactions with the `depth` query parameter and the by-field filtering. As the `depth` parameter was no longer being used on the V2 API (all nested FKs are now handled by their own Serializer), I ended up removing this functionality. Once this was done, filtering worked perfectly.

Once this update was done, I had a few N+1 errors to fix due to the new Serializer required to allow full removal of the depth parameter processing.

## Related Issue

Closes #719 

## How was this tested

First, the change was A/B tested against `staging`. Next, a Django test server was set up and I took a good poke around, experimenting with passing different query parameters, etc.

Because this PR changes the shape of the API response for some endpoints, some of our tests needed to be updated. Once this was done all tests were passing when Pytest was run


augustjohnson (approved):
Love it, excellent change!

PR #730, opened 2025-05-16
[V2] Weapon Properties refactor

### LINKED ISSUE #710: Revisit Weapon Properties fields
## Original Feedback

The way that this model handles a weapon's properties is somewhat confusing currently.

There is a "properties" array of strings of weapon properties. There are also a series of boolean fields with names like "is_heavy", "is_simple", and sometimes property data is repeated. I think we should pick one approach and stick to it.

Would it be easier to create a new data model for weapon properties?

Example:

```json
            "url": "https://api-beta.open5e.com/v2/weapons/srd_blowgun/",
            "key": "srd_blowgun",
            "is_versatile": false,
            "is_martial": true,
            "is_melee": false,
            "ranged_attack_possible": true,
            "range_melee": 5.0,
            "is_reach": false,
            "properties": [
                "ammuntion (range 25.0/100.0)",
                "loading"
            ],
            "distance_unit": "feet",
            "name": "Blowgun",
            "damage_dice": "1",
            "versatile_dice": "0",
            "reach": 5.0,
            "range": 25.0,
            "long_range": 100.0,
            "is_finesse": false,
            "is_thrown": false,
            "is_two_handed": false,
            "requires_ammunition": true,
            "requires_loading": true,
            "is_heavy": false,
            "is_light": false,
            "is_lance": false,
            "is_net": false,
            "is_simple": false,
            "is_improvised": false,
            "document": "https://api-beta.open5e.com/v2/documents/srd/",
            "damage_type": "https://api-beta.open5e.com/v2/damagetypes/piercing/"
```

calumbell:
## Description

This PR refactors the way that `Properties` on the `Weapon` model are handles by the API. The new `WeaponProperty` model is used to capture different Weapon Properties, removing the needed to use a whole stack of booleans on the `Weapon` model to encode that weapons properties.

Currently on the staging

**Previously**
```
  "url": "https://api-beta.open5e.com/v2/weapons/srd_blowgun/",
  "key": "srd_blowgun",
  "is_versatile": false,
  "is_martial": true,
  "is_melee": false,
  "ranged_attack_possible": true,
  "range_melee": 5.0,
  "is_reach": false,
  "properties": [
      "ammuntion (range 25.0/100.0)",
      "loading"
  ],
  "distance_unit": "feet",
  "name": "Blowgun",
  "damage_dice": "1",
  "versatile_dice": "0",
  "reach": 5.0,
  "range": 25.0,
  "long_range": 100.0,
  "is_finesse": false,
  "is_thrown": false,
  "is_two_handed": false,
  "requires_ammunition": true,
  "requires_loading": true,
  "is_heavy": false,
  "is_light": false,
  "is_lance": false,
  "is_net": false,
  "is_simple": false,
  "is_improvised": false,
```

The new format takes many of the `"is_something"` boolean fields on the `Weapon` model and presents that data through the `properties` array. This field contains an array of FKs to the `WeaponProperty` model. Long term this is a much more flexible and maintainable system.

```

{
    "url": "http://localhost:8000/v2/weapons/srd_blowgun/",
    "key": "srd_blowgun",
    "properties": [
        {
            "property": {
                "name": "Ammunition",
                "url": "http://localhost:8000/v2/weaponproperties/srd-2014_ammunition-wp/"
            },
            "detail": "range 25/100"
        },
        {
            "property": {
                "name": "Loading",
                "url": "http://localhost:8000/v2/weaponproperties/srd-2014_loading-wp/"
            },
            "detail": null
        }
    ],
    "distance_unit": "feet",
    "name": "Blowgun",
    "damage_dice": "1",
    "range": 25.0,
    "long_range": 100.0,
    "is_simple": false,
    "is_improvised": false
}
```

Note that I haven't touched some of the fields which one might expect to be refactored as `WeaponProperties`. This is because some are properties in the strict sense (ie. whether a weapon is simple/martial or improvised is not listed in the Property column of the Weapons table in the 5.1 WotC SRD). 

Effort was taken to preserve the filtering options from the previous version of the `v2/weapons` endpoint, as requested in a comment on the related issue. These could easily be expanded.

## Related Issue

Closes #710 

## How was this tested?

First, the the changes were visually inspected using the Django REST Framework HTML view of the API response to confirm that the endpoints were working as desired. After running the smoke tests locals, `pytest` was run, failing several tests. These were inspected and found to be due to the desired changes in the shape of the API response taking effect, and the tests were updated to reflect this.

## Additional Context

A big part of the motivation for getting this issue resolved was to get the API ready to add SRD 5.2 weapons to O5e by providing a way to handle Weapon Mastery Properties. The new system could well handle this in its current state, but we will probably want to add a way to distinguish between regular and weapon matery properties - perhaps by added a `type` field to the `WeaponProperty` model?


api_v2/tests/responses/TestObjects.test_item_ranged_weapon_example.approved.json
```diff
@@ -61,35 +61,28 @@
...
-            "two-handed"
+            {
+                "detail": null,
+                "property": {
+                    "name": "Two-Handed",
+                    "url": "http://localhost:8000/v2/weaponproperties/srd-2014_two-handed-wp/"
+                }
+            },
+            {
+                "detail": "range 150/",
```
augustjohnson, line 78: This seems like possibly a bug, how is long range represented? It's without a doubt an improvement from text though.

api_v2/models/weapon.py
```diff
@@ -3,13 +3,39 @@
...
+class WeaponPropertyAssignment(FromDocument):
+  """
+  This is an intermediate model that is used to assign WeaponProperties to
+  Weapons while bundling in any extra contextual data in the `detail` field
+  """
+  weapon = models.ForeignKey(
+    'Weapon',
+    related_name='properties',
+    on_delete=models.CASCADE
+  )
```
augustjohnson, line 31: Should we consider adding a pretty-print of the properties like the documentation provides? It should be fairly simple to concatenate.

augustjohnson (approved):

api_v2/tests/responses/TestObjects.test_item_ranged_weapon_example.approved.json
calumbell: Good spot, this is a typo. Should read `"range 150/600"`.

I didn't remove the Range fields from the `Weapon` model because I wasn't fully satisfied with encoding the range/long range using a string. It does allow us to reconstruct verbatim the source material, but it feels like this data shouldn't be so deeply buried.

PR #735, opened 2025-05-17
Convert missing v1 data to v2

eepMoody:
## Summary

Doing some experiments here with GPT Codex, to see if it can streamline the tedious work of converting between versions. Thus far it appears promising.

This PR does the following:

- add script to convert v1 JSON to v2
- create v2 feats and subclass data for Tome of Heroes and Tal'dorei
- add open5e backgrounds and subclass data
- include Shade species and Stoor Halfling subspecies
- ensure tests and quicksetup succeed

## Testing
- `pipenv run python manage.py quicksetup --noindex`
- `pipenv run pytest`


eepMoody:
@augustjohnson I _think_ this is gtg now, based on some local testing it appears to be behaving, and all the new data appears to be present. Want to give it a once-over and make sure I/it did everything right?

augustjohnson (changes requested):
Excellent progress. A few questions.
1. Can you put convert_missing.py into the scripts / data manipulation folder?
2. Did you review the a5e mismatches? I can look into it if you haven't, but there's a mismatch there.
3. What was the additional ToH background? v1 has 18, v2 has 19.

eepMoody:
@augustjohnson 
- Moved the script!
- The background appears to be "Scoundrel", which was included in the v1 "o5e" backgrounds despite appearing to originate in Tome of Heroes. Looks like the AI caught it and fixed it while reviewing the content. Smart robot!
- The extra a5e feats are real, but it looks like we were using a source for that that included some non-srd items. I've removed them.
- The other mismatches in A5e I'm having trouble finding. I see all the same race/species as I'm spot-checking, but they're organized differently because of the new structure. Same for classes/subclasses. I thought maybe we had the same SRD issue, but I'm not seeing it. Maybe they're just getting counted different because of the structural changes?

augustjohnson (approved):

Issue #711, opened 2025-04-23
Ingest 2024 SRD data


eepMoody:
Data from the new 2024 SRD ([available here](https://www.dndbeyond.com/srd) should be converted to JSON and ingested.

- [x] Creatures
- [x] Classes
- [x] Subclasses
- [x] Items
- [x] Spells

PR #748, opened 2025-05-26
Ingest 2024 Spells, Monsters and Items

eepMoody:
# SRD 2024 Data Import - Part 1

## Suggested review focus
This is obviously massive, so I suggest we focus our review as follows:
- Check changes to model
- Check infrastructure changes
- Spot check new items to ensure there were no systemic conversion errors
- **Ignore all the markdown files & conversion scripts.** They are very much bespoke franken-scripts made with AI, liberal copy-pase, and a bunch of trial and error.
  - They're also bespoke to the task and don't need to work outside of the product they produced. I'm mostly keeping them here in case they are useful to someone in the future.

## Major changes
- Imported all raw source files from SRD 5.2
- Wrote bespoke conversion scripts for monsters, animals, classes, spells, and equipment
- Added "type" to weapon properties to support new 2024 "mastery" rules

## Related infrastructure Improvements:
- Updated standard django test command to use pytest by default
- Enhanced quicksetup command for better development workflow

## Remaining work
- Character origins, backgrounds, and related objects
- Rules sections

## Why stop here?
This PR focuses on a set of relatively isolated game objects that don't have complex dependencies (such as character classes, which I intend to tackle last).

These items are also relatively similar to their 2014 counterparts. Follow on PRs will tackle the more unique items that will require more difficult decisions about API structure, but I wanted to get this basic stuff in there first so we had a starting point.

makes progress on #711


data/v2/wizards-of-the-coast/srd-2014/ItemSet.json
```diff
@@ -41,7 +41,24 @@
       "srd_smiths-tools",
       "srd_tinkers-tools",
       "srd_weavers-tools",
```
eepMoody, line 43: In the case of these category-like itemsets, I thought it best to include both versions of each item. That way the result is good if the user has filtered to one ruleset, and there isn't a "wrong" tools set to filter by sitting around as a footgun.

data/v2/wizards-of-the-coast/srd-2014/Rule.json
```diff
@@ -1933,7 +1933,7 @@
 },
 {
   "model": "api_v2.rule",
-  "pk": "srd_mounts-and-vehicles_tack-harness-drawn-vehicles",
+  "pk": "srd_mounts-and-vehicles_tack-harness-land-vehicles",
```
eepMoody, line 1936: I generally renamed "drawn" vehicles to "land" vehicles in our pks, to match the standard language and proficiency used for them elsewhere. I figure this is more user-friendly and expected even if a header calls them something else in the doc.

api_v2/management/commands/import.py
```diff
@@ -17,6 +17,9 @@ def add_arguments(self, parser):
                             "--dir",
                             type=str,
                             help="Directory to write files to.")
+        parser.add_argument("--debug",
```
eepMoody, line 20: I've added this to improve error logging. Now it should be possible to see which file and line caused the import to fail

api_v2/management/commands/import.py
```diff
@@ -28,5 +31,151 @@ def handle(self, *args, **options) -> None:
             exit(0)
 
         fixture_filepaths = glob.glob(options['dir'] + '/**/*.json', recursive=True)
+        
+        self.stdout.write(f'Found {len(fixture_filepaths)} fixture files to load.')
```
eepMoody, line 35: This section modifies the import flow in the following ways:

- Integrity checks are held off while fixtures are loading, to allow circular dependencies to succeed. Now necessary due to new relationships between SRD 2014 and 2024
- After all data is loaded, integrity checks are run
- All errors (not just one) are printed to the terminal in case of failure
- This now includes both the item and nature of the error for each issue, to speed up debugging.

data/v2/wizards-of-the-coast/srd-2024/CharacterClass.json
augustjohnson, line 1: No artificer in the SRD.

augustjohnson:
Here's notes on items.

- [ ] Within the old SRD import, I broke magic items into their actual usable versions, for example, instead of Adamatine Armor, I created: https://api-beta.open5e.com/v2/items/srd_adamantine-armor-breastplate/. I would like to continue this pattern, but it's probably manual work.
- [ ] The mapping from item to weapon seems to be good to go! Nice work on that, same for armor.
- [ ] Service seems like an awkward addition to items, simply because Item inherits from Object (meaning we are required to include a weight for a given service now). I'd prefer to put it elsewhere (new model?).
- [ ] It's missing a large subset of mundane items. (Bell, Bedroll, Chain, Caltrops, Iron Pot, etc). These are around page 100 of the SRD. I did import them on the old version: https://api-beta.open5e.com/v2/items/srd_pot-iron/.

calumbell (changes requested):
Nice work! Exciting to finally have material from the 2024 SRD to the dataset!

# Requested Changes

The changes I think we should make before we merge this PR are as follows:

1) Make sure the new PKs are consistant: most begin with `srd2024`, but this should be `srd-2024` to better match existing naming conventions. ie. the 5.2 Aboleth has a PK of “srd2024_aboleth”, but this should be `srd-2024_aboleth.

2) Remove the Artificer from `data/v2/wizards-of-the-coast/srd-2024/CharacterClass.json` (as per August’s comment).

2) Add the SRD 5.2 Fighter to `data/v2/wizards-of-the-coast/srd-2024/CharacterClass.json`

W.R.T. the other points flagged by @august, I think that the following are better served by seperate own issues, as they require some design work that puts them out of the scope of this PR:

- Sorting out weirdness with Services inheriting from the Object model

The following points feel perfectly appropriate to touch on in this PR. But they are simple enough to add in later, and marginal enough that it is probably isn’t worth holding up merging this PR to `staging` on their account.

- Missing mundane items
- Generate ‘usable’ magic items from templates (ie. Adamantine Armor)

## Models Changes 
Doesn’t look like you needed to update any models, let me know if i am being a numpty.

## Spot Checking

`v2/creatures/srd2024_aboleth`

This was compared to the 2014 SRD version on O5e (`/v2/creatures/srd_aboleth`) and the Aboleth statblock from the 5.2 SRD.

The JSON looks good to me. Every field in the new statblock is represented in the API. Actions now appear to be sorted by their original order within the statblock and by the action type – which is a nice touch.

One issue is that the primary key doesn’t match established convention on how we handle year suffixes (ie. `tob-2023`). The `srd2024` should be `srd-2024`


`v2/weapons/srd2024_battleaxe/`

Looks good! Comparing it to the SRD PDF it looks like it captures everything.

As above the key should be `srd-2024_battleaxe`


`v2/items/srd2024_bag-of-tricks/`

Captures everything in the source. Again, the `key` is incorrect. It should be `srd-2024`.

The fields relating to AC, resistances, and size seem superflurous to me, but this isn't the PR to be changing this on.

`v2/spells/srd2024_magic-missile/`

Seems to capture everything in the source. The `casting_options` field is an empty array, which isn't the case with the equivelent 5.1 SRD entry (`/v2/srd_magic-missile`) Again the key is incorrect.

`v2/classes/?document__key__in=srd-2024`

Looks good, but the class list is a bit wrong. As has already been mentioned, the Artificer shouldn’t be here. I also clocked that the SRD 5.2 Fighter is missing.

The `features` array for SRD 5.2 classes, but your PR states that classes will be handled at a later time.

Again, a missing hyphen in the `key`s

## Check Infrastructure Changes

Eyeballing the code, the infrastructural changes look sound. Deleting the DB and re-running `quicksetup` appears to work exactly as intended. `pytest` also running as expected.

eepMoody:
@calumbell  Cool. I'll get started on these!

As for the classes, they are minimally stubbed out intentionally. We need some to _exist_ because some of the items being imported reference them. Specifically some spells and a few items—which is why I'm not including Fighter yet, it's not needed. I'll go ahead and delete artificer to keep it clean.

I want to tackle classes properly at the end because of how deeply-connected to many of the other objects they are, and we can add fighter when we need them IMO.

Full list of intended corrections:

- [x] Globally fix key prefix to `srd-2024`
- [x] Remove Services for now
  - [x] Create issue to add them back in once we figure out how best to handle them (#751)
- [x] Create issue to populate casting options, once #749 is merged in  and we've fully resolved how we're handling those (#753)
- [x] Remove Artificer from class list
- [x] Add mundane items from p100 of the SRD
- [x] Create issue to break down "usable" magic items (#752)
  - I want to tackle that separately since it requires making some editorial choices about the raw SRD content, and a separate PR will provide a better place to review and discuss the results.

calumbell:
@eepMoody looks good! Ping me when things are ready and I will review again

eepMoody:
Okay @calumbell I _think_ this is good to go. Which given the goal of merging existing stuff to `main` soon leaves the question of what's best to do with all this.

- Merge it in, and just release an incomplete slice of the 2024 rules for now
- Create a feature branch and change this PR to merge there instead so we can get the rest done in one cleaner place
- This IS the feature branch, start cranking through other properties (I think this might get a lil messy but that's fine I guess)

eepMoody:
I also went ahead and took a first pass at spell levels. I don't think they're 100% yet, my spot-checking shows there are still some issues.

But I'm not seeing anything that's _wrong_ or _missing entirely_ in my spot checks at this point, just things where there's more detail to be added to the casting objects. Since those details are already in the description, I'm going to call this one "done" for now.

To handle some truly bizzare or conditional ones I went ahead and added an optional `desc` field to the spellcasting option model. That gives us a place to put things like:

- "with a 9th level slot, can modify a memory up to 366 days in the past"
- "with an 8th level slot, can animate 4 ghasts, 2 mummies or wights, or 1 vampire"
- "increases the number of gallons created by 2, or the size of the cube of rain by 5 ft"
- "use spell level 8 where indicated in the stat block"
- "blocks spells of up to level 6"

And so on. Where possible, I tried spread the calculations and have the casting option enumerate that specific level's effect.

**Testing done**

- Ran a script to verify that any spell with a "Higher Level" description has at least one "meaningful" (non-null) spellcasting option, a good heuristic that we've captured changes for all the spells that actually need them
- Spot checked about 15% of the total corpus, last two checks ran clean with only minor details omitted from desc

calumbell (approved):
Looks good! Spot checking the data and it looks like the things i'd flagged previously are all fixed. Still plenty left to do, but I reckon we merge your work so far to keep the PRs managable. Nice one!

PR #749, opened 2025-05-31
706 casting objects returns null if no values

### LINKED ISSUE #706: Casting Objects returns null if no values
`"casting_options"` should only be returned if they contain any interesting information. Some options return an object where all the keys are `null`

Based on #687

augustjohnson:
This PR removes all empty casting options. This should result on only presenting default casting options, ritual casting options, and casting options that specifically modify a spell's effect or requirements.


eepMoody (approved):

eepMoody:
Nice! :shipit:

PR #750, opened 2025-06-01
Add fuzzy and vector search to v2/search endpoint

eepMoody:
## **Improvements:**
- **Fuzzy search** using `RapidFuzz` for typo-tolerant queries ("birb" → "Bird") against names only
- **Vector search** using `scikit` TF-IDF for semantic matching across text

## **Docker change:**
This now uses a python `3.11.1-slim` environment instead of `3.11.1-alpine`. `slim` comes with some improved C and python-wheel compatibility required for most well-supported vector search libraries. It adds a bit of size to the OS (~20mb) but is still very workable in DO Apps.

**Tested live in DO to confirm. No issues. No visible changes to performance, cost, or resource usage (aside from expected ~20mb increase in disk usage, which is not impactful)**

## **New default search behavior**
- By default, runs an exact query
- If no exact results found, runs a fuzzy name-only search (slower) and returns that instead
- If the user requests vector or fuzzy search they are added to exact methods
- `strict` mode allows the user to specify exactly which search methods to use and ignore the fallback logic
- If multiple modes are requested, they are merged and de-duped in order of decreasing precision:
  1. exact (name field)
  2. exact (other fields)
  3. fuzzy (by similarity, descending)
  4. vector (by similarity, descending)

## **Performance**
All searches perform within a reasonable amount of time, and appear to be improved over existing search.

The fuzzy fallback method is still quite performant, but appears slower because it runs sequentially after a failed exact search.

### Search Mode Performance Comparison (local machine)

| Query | Default Mode | | Fuzzy-Strict | | Vector-Strict | |
|-------|-------------|---|-------------|---|-------------|---|
| | Time (ms) | Results | Time (ms) | Results | Time (ms) | Results |
| fire | 185.5 | 653 | 140.9 | 13 | 170.0 | 32 |
| sword | 149.3 | 133 | 111.7 | 19 | 100.6 | 18 |
| heal | 127.0 | 49 | 113.4 | 20 | 101.2 | 16 |
| dragon | 196.6 | 363 | 141.3 | 3 | 128.4 | 20 |
| fireball | 108.2 | 34 | 82.7 | 1 | 66.9 | 4 |
| teleport | 162.8 | 116 | 81.7 | 0 | 108.4 | 22 |
| magic weapon | 173.0 | 622 | 106.4 | 0 | 253.6 | 1372 |
| chromatic orb | 32.5 | 1 | 75.0 | 1 | 123.2 | 34 |
| firbal (fuzzy fallback test) | 121.5 | 4 | 86.4 | 4 | 10.9 | 0 |
| dragn (fuzzy fallback test) | 183.9 | 157 | 188.0 | 157 | 9.7 | 0 |

### **Dependencies Added:**
- `rapidfuzz` for fuzzy string matching
- `scikit-learn` for search vectorization
- `numpy` for maths
- `python 3.1.1-alpine` -> `python 3.1.1-slim`


augustjohnson:
Some general comments: 
I'm allergic to dependencies so adding 3 makes me unhappy. Similarly for switching to debian-slim, 20mb of what? (probably not C compatibility). Not deal breakers, and I understand it wont' work without them, but we just added millions of lines of OPC (other people's code) that we are now responsible for maintaining and that's best avoided.

Beyond that, I'm very happy with the tests being added to search, and the refactoring of existing seems like a net positive for sure. 

We may want to stress test it to see if we can blow up the threads and crash it. 

Overall, the CODE changes are almost certainly superior and just fine. I do want to get the UI out there so I can see if it's actually BETTER or not.

calumbell (approved):
Looking over the code, the quality looks good. The docstrings are very helpful and methods/variables have sensible names, so that I was able to follow the logic easiliy.

A few more comments wouldn't hurt, and there might be some scope for refactoring to the `SearchConfig` class in `search/apps` to reduce nesting, but this is me nit-picking.

## Testing

`quicksetup` runs without error. Running `pytest` against the local Django server turns up no failed tests. I ran a few queries and the fuzzy search appears to be working precisely as described.

It feels much faster also!! I spun up a Nuxt test server to see how the new change played with the front-end and it is much, much snappier. Nice work!

calumbell:
@augustjohnson Didn't see your comment until after I had finished my review - apologies!

Good point about dependencies. Can't say I have a problem adding numpy as a dependency, it is a mature and well-supported package, might be worth holding off on merging until we have investigated  the others?

eepMoody:
So as far as the dependencies, I did try to find the best, most well-supported options I could. And I took the approach of finding the most isolated and minimal options to solve each need. There's not really any way to do these things ourselves: or at least, I think it's very complex specialized stuff that's outside the scope of this project.

Search is IMO like crytpography: rolling your own is almost always a mistake, no matter how simple you think it'll be.

## Numpy

https://github.com/numpy/numpy 

Numpy is _the_ python data science library. I was legitimately surprised we didn't have it already.

Here we are using it primarily for its N-dimensional array framework and linear algebra libraries, which are fundamental to any form of vector search or ML training.

## Scikit

https://github.com/scikit-learn/scikit-learn 

SciKit is a standard-issue ML library. It is well-maintained FOSS software (updated today!) and very specialized for this kind of work. Alternatives are more fully-featured search libraries, but the more I looked around it seemed better to keep this simple. We could also move up to something like pytorch or tensorflow, but they're really overkill for our current state. (maybe someday!)

It is also low-risk and isolated, in that the vectorized pickle file it creates is interchangeable with the output of any similar library. If it dies we can just swap it out and do some light updates to handle whatever format changes come with it—though I will likely be distracted, as the global science and data industries will in crisis mode.

It is under the 3-Clause BSD license (even more permissive than MIT) so we're good on that front.

## Rapidfuzz

https://github.com/rapidfuzz/RapidFuzz

RapidFuzz is another fairly common-use python library. It is an MIT-licensed fork of [FuzzyWuzzy/theFuzz](https://github.com/seatgeek/thefuzz) (which at one point went GPL Licensed, inspiring this fork). It is actively maintained and widely-used. 

> _Sidenote: Hilariously, this is now a dependency of the same project it was originally forked from, which went back to the MIT license eventually 😆. The Fuzz is also a viable alternative, but is more complex and maintained by SeatGeek so I went with the the simpler, FOSS option._

Again, I think this is fairly isolated and the most viable alternatives are more deeply integrated Search frameworks, which felt riskier to me. If this stops getting updates, there are other forks of FuzzyWuzzy, and other fuzzy algorthims we could replace it with. Currently, I don't think we're going to find many lighter options that deliver this kind of performance.

##  Python slim

`python 11.1-slim` (based on debian-slim) is an Official Python™ Image, on the same [list](https://hub.docker.com/_/python) as `alpine` our previous image. It is a more complete but minimal debian OS environment with better package/wheel compatibility (vs Alpine which is a truly barebones environment) and is intended for exactly this sort of hosted project. The 20mb of extra code are primarily standard linux/debian modules to flesh out the OS. It admittedly includes a lot of core OS stuff we _don't_ need, but it's really the "next step up" in terms of industry-standard python images.

### Alternative

~~The most obvious alterative is to start bundling in all the dependencies required to compile our other libraries using C and python wheel frameworks, but that starts to move into the "build and maintain your own image" territory. I was about 15 python dependencies deep before I stopped and checked to see if there was a better image we should be using. There would definitely be more things for us to manage if we went this route.~~

EDIT: On further investigation, the _core_ issue is that alpine uses [musl libc⁠](https://musl.libc.org/) instead of the more common [glibc](https://www.etalabs.net/compare_libcs.html), which causes generalized incompatibility across the ecosystem. So that dependency chain would not really have ended I think until I had more or less replaced the entire image.

We could theoretically try to cobble something together out of non-standard libraries that will run on Alpine. But it's slim pickings and I think the result will be very ugly and brittle.

Or we could use Slim to build, then switch to Alpine to run, which would work but now we're dependent on _two_ images which I don't love.

Last would be to just not add this kind of functionality at all, but I don't think that's the right call.

### Stability

It's actively supported (last update a [couple weeks ago](https://hub.docker.com/layers/library/python/3.11-slim/images/sha256-7ad180fdf785219c4a23124e53745fbd683bd6e23d0885e3554aff59eddbc377)) by Python, at the same level as Alpine.

### TL:DR;

I did a lot of digging trying to avoid it, but there aren't a lot of these sorts of data-intensive libraries that will run on something as barebones as Alpine. 

We can theoretically make it work, but we would effectively be switching to worse-supported, fringe ML and search libraries to avoid using a stock-standard image, which seems like the wrong call to me.

eepMoody:
@augustjohnson @calumbell writeup on the dependencies above! Didn't take very long, I was taking a fairly skeptical approach myself so I already had all the info in my browser history 😆

eepMoody:
Also just a thought on the original choice to use Alpine that came out of the above investigation: I'm not sure it was ever actually the best choice for us.

From the Python docs:

> [Alpine] is useful when final image size being as small as possible **is your primary concern**. The main caveat to note is that it does use [musl libc⁠](https://musl.libc.org/) instead of [glibc and friends⁠](https://www.etalabs.net/compare_libcs.html), so software will often run into issues depending on the depth of their libc requirements/assumptions.

IMO minimum size isn't actually one of our primary concerns. DO has plenty of space, and we're nowhere near memory or disk constraints. If we were trying to run this as a distributed server on a bunch of Raspberry Pis or something I think I'd lean towards it, but package size on the order of MB really doesn't affect any of our use-cases.

I _think_ our main concern on this axis is standardization and simplicity, which alpine actually moved us away from.

Here we're hitting on a cost of that, where industry-standard libraries for things like search, math, and data science are incompatible with our base image.

The more I read up on the choice of images, the more I think we should actually consider moving from Slim to the standard Python image (later, as a separate project, with extensive testing) to even further simplify and standardize our build.

calumbell:
@eepMoody Thanks for taking the time to write this up! Your reasoning is solid, and you have already considered escape hatches if the dependencies stops working. I am happy with these changes, what are your thoughts @augustjohnson ?

augustjohnson:
My thoughts are MERGE!

PR #755, opened 2025-06-04
Add Services to API

### LINKED ISSUE #751: [TODO] Add Services to API
## Description

These are treated as items within the SRD, but as pointed out in review of #748 they are awkward within our model. We should likely either:
 
- Figure out how to add them as items (make a few irrelevant fields nullable?) 
- Create a new object to handle them

## Requirements

- [ ] All services from 2024 and 2014 SRD are ingested and added to the API

calumbell:
## Description

This PR adds the `Service` model for handling those purchases able services / hireling expenses which don't neatly fit in the `Item` model (with their `size` and `weight` fields). It additionally adds serializers and views so that Services can be accessed via the `/v2/services` endpoint.

## Related Issue

Closes #751 

## How was this tested?
- Rebuilding the DB to make sure that everything ran as expected.
- Visiting `/v2/services` on the feature branch to confirm that Services were returned
- Running `pytest` to confirm that all tests were still passing.

## Additional Context

There is a good chance we will want to integrate this model with the `Items` in our dataset because they are conceptually closely linked. I opted not to handle this just yet pending further discussion with other developers w.r.t. how we wanted to handle this.


augustjohnson (changes requested):
Add data for Hirelings and Spellcasting Services

calumbell:
@augustjohnson Thanks for flagging that those were missing. They are now in the dats set!

augustjohnson (approved):

PR #758, opened 2025-06-07
Create shared concepts construct for conditions, damage types, etc.

eepMoody:
Currently this is a POC that I will flesh out based on discussion. 

It adds a higher-level "conditions_concept" object that spans across gamesystems. This collects all condition definitions and merges those with shared names to create a common object with all defintions, keyed to gamesystem.

The idea is that a frontend consumer could hit this endpoint for a condition without knowing what gamesystem they want to use (or even what conditions exist in each), and then create their own fallback logic (given the user's selections) to which game system they want to use, and in what order.

Rough concept example (without having yet dug into the summary values that I should expand these links into).
If this works I'll add a gamesystem key and other details to help the user distinguish which one to use.

```
{
    "name": "Invisible",
    "key": "invisible",
    "url": "http://localhost:8000/v2/condition-concepts/invisible/",
    "desc": "The invisible condition as it appears across different game systems: 5th Edition, Advanced 5th Edition.",
    "conditions": [
        {
            "name": "Invisible",
            "key": "a5e-invisible",
            "url": "http://localhost:8000/v2/conditions/a5e-invisible/"
        },
        {
            "name": "Invisible",
            "key": "invisible",
            "url": "http://localhost:8000/v2/conditions/invisible/"
        }
    ]
}
```


augustjohnson:
I think the idea makes sense. So using damage type as an example, a shortsword would have the DamageTypeConcept: "Slashing" linked, and slashing would then have srd_slashing linked as well. So there'd be some sort of generic description, and filters would all work on the concept, but given something specific, there'd be a detailed description.

I like it.

Data model objects that could be considered for generics:

- [ ] Ability
- [ ] Skill
- [ ] Alignment
- [ ] Armor
- [ ] Weapon
- [ ] Condition
- [ ] DamageType
- [ ] Environment (already kind of generic? It's not a first-tier model in the SRD)
- [ ] ItemCateogry
- [ ] ItemRarity
- [ ] Language
- [ ] Size

eepMoody:
@augustjohnson Yeah that's exactly how I'm thinking about it!

I'll flesh it out a bit more and see if it holds up, then we can take a proper look down the road.

I _was_ going to steer clear of things like items and weapons for now, just because they're already quite complex and I think will end up being too tightly-coupled to the rules to be truly portable. I'm also not sure if I've found any distinctions for things like Abilities, Skills, and Alignments that would justify adding the top-level concept.

But I guess the nice part about it being a non-breaking way to add a grouping concept is that it would be easy to add when it turns out I'm wrong about one of those!

PR #761, opened 2025-06-08
760 bug data files not organizedimportedexported correctly

### LINKED ISSUE #760: [BUG] Data files not organized/imported/exported correctly.
## Bug description
When a clean checkout runs quicksetup, it vacuums in all data in the /data directory. Then if they run export -d ./data it should blow out all of the same data, meaning basically no changes. 

This does not appear to be the case currently, various fields, documents, and keys are changes and reorganized in a way that does not make this process symmetrical.

## Steps to Reproduce
clean checkout
quicksetup
run manage.py export -d ./data

augustjohnson:
This PR catches the /data directory up with the current models. All fields should now be explicit. All files should now be properly documented. This change should not functionally make any difference, but the edits in the future will be much more obvious and less likely to end up changing dozens or hundreds of lines.


eepMoody:
I know this is a pretty rapid review for _checks notes_ >100k lines of data, but this looks solid to me!

Focused on examining the test cases and ensuring that nothing was getting bulk-deleted, which would be the real scary thing!

eepMoody (approved):

eepMoody (approved):
Sorry, ran into some wokyness that turned out to be from cached stuff. Blew away my local env tested one last time, LGTM!

PR #767, opened 2025-06-15
676 add monster illustrations as a field to creature objects

### LINKED ISSUE #676: Add monster illustrations as a field to creature objects
v1 monsters had a few instances where they have beautiful art. We need to support this within v2.

Add an "illustration" field which will link to an "image" object.
Load the relevant data into the correct directories in the repo.

augustjohnson:
This adds an illustration image field to creatures, and includes mapping between the illustrations created by Violet, and the creatures (in the SRD) that need them.


calumbell (changes requested):
Looks good - really nice to get this functionality added back in to the API.

Requesting changes because I am not sure that the `/server/vector_index.pkl` should be a part of this commit. This should probably be added to the .gitignore?

eepMoody:
Yeah it probably should be part of the gitignore. But it's also pretty harmless. The server build process will blow it away anyways, so we can probably just let this go as-is and follow up to clear it out?

calumbell (approved):
@eepMoody Yeah it isn't really a big deal, it just sometimes gets in the way when I switch branches. In any case it looks like it has already been added to the `.gitignore` on the `staging` branch so we are good to go!

PR #777, opened 2025-07-25
Update 5.2 srd magic items

mshea:
I added tables to a bunch of 5.2 SRD magic items that were missing them. I don't know why the diff is so off...


mshea:
You can also ignore my gitignore. I don't know why that got pushed either. I had some local temp files to aid in this that I no longer need.

calumbell:
Nice, thanks @mshea! It looks like the diff went bananas because a level of indentation was added to the JSON between the opening and closing square brackets. I'll push a commit to this PR that brings the diff down to +43 -40, will be much easier to review like that!

calumbell (changes requested):
Thanks for getting this fixes submitted, it wasn't even on my radar that these tables were missing!

There is one minor issue with some of the tables. In some cases longer tables that contain very short pieces of data are rendered as two columns in the SRD 5.2.1. For example take the Bag of Tricks (pp. 212).

Screenshot 2025-07-25 at 14 48 46

It is one table, but the column count has been double to better utilise the horizontal space. If we look at how the table on the Open5e site when it consume this data:

Screenshot 2025-07-25 at 14 31 02

It looks kind of goofy. It would be great if we could preserve that double-column layout!!

Would it be possible to refactor the following items so that the double the number of columns and halve the number of rows?

- Ammunition of Slaying (srd-2024_ammunition-of-slaying)
- Armor of Resistance (srd-2024_armor-of-resistance)
- Bag of Tricks (srd-2024_bag-of-tricks)
- Candle of Invocation (srd-2024_candle-of-invocation)
- Dragon Scale Mail (srd-2024_dragon-scale-mail)
- Potion of Resistnace (srd-2024_potion-of-resistance)

Everything else is perfect, no notes. Thanks so much for getting these changes submitted!

augustjohnson (approved):

augustjohnson (approved):

PR #783, opened 2025-08-17
Adding icontains query for species name.

### LINKED ISSUE #779: [BUG] V2 Species / Backgrounds / Feats / Rules endpoints not filterable by `name__icontains`
## Bug description
Filtering the `/v2/species` endpoint using a partial match against the `name` field via the `?name__icontains` query parameter has no effect on the API results returned by the endpoint. This query parameter should allow users to filter the results.

## Steps to Reproduce
1. Spin up a Django test server from  the `staging` branch of the API.
2. Vists the `/v2/species` endpoint, note that 54 species are returned.
3. Now try filtering by `name__iexact`, case-insensitive exact match against the value of the `name` field. Visiting `/v2/species/?name__iexact=dwarf` returns 1 result, the `srd-dwarf`. So far so good.
4. Now try filtering by `name__icontains`, case-insensitive partial match. Visiting `/v2/species/?name__icontains=dwarf` returns 54 results. It should be at least 1 (`srd_dwarf`, and whatever else). No filtering is taking place

augustjohnson:


augustjohnson:
One-line change to add in icontains in Species Name. Did not impact tests, but I tested the scenario in the bug along with elf, human, gnome. Seemed to come back with reasonable results.

PR #788, opened 2025-08-28
Updated 2024 spells Wish and Fireball

mshea:
Updated Wish and Fireball with appropriate text


calumbell (approved):
Hi @mshea, thanks for getting this fixed! I added an extra commit to update the line breaks to match our house style, otherwise it all looks good to me!

Issue #791, opened 2025-09-04
[TODO] Deprecate "Condition Concept"


augustjohnson:
## Description

We should deprecate the Condition Concept out of v2, it has been conceptually replaced by "core" and various add ons.

## Requirements

*How do we know it's done?*
- [ ] Remove condition concept model
- [ ] Remove condition concept serializer and view
- [ ] Double check all tests
- [ ] Confirm no data exists.

Issue #790, opened 2025-09-04
[DATA] Add in a5e core concept descriptions.


augustjohnson:
Add in a5e descriptions for core concepts.

In chapter 7: Ability Scores
- [x] Ability Descriptions
- [x] Skill Descriptions

In Chapter 19.1 Monsters A-F
- [x] Creature Types

No reference found? Maybe ignore these?
- [ ] Alignment Description
- [ ] Damage Type Description

Issue #793, opened 2025-09-04
Core concepts applied to Conditions


augustjohnson:
The core concepts that apply to Damage Type, Creature Types, etc should also apply to conditions.

- [x] Move conditions in Core document.
- [x] Modify model to allow for descriptions into their own documents.
- [x] Add descriptions from srd-2024.
- [x] Add descriptions from a5e.
- [x] Update tests

PR #794, opened 2025-09-04
461 determine a structure for equivalent concepts across systems

### LINKED ISSUE #461: v2 feature: equivalents
For a given model, include an "equivalents" many-to-many relationship where a given object can give a list of bi-directionally related to "equivalent" objects from different documents.

Size-Large from the SRD is technically distinct from Size-Large from the A5e SRD, but they are equivalents.

augustjohnson:
Adding in significant conceptual changes to shared concepts like "Acid Damage".

From a high level perspective, a subset of data is considered canonical to 5e, and when a creature from creature codex references "perception check" or "thunder damage" they fundamentally mean the same thing as the other documents. This will allow for shared filtering.

The data that this PR changes to core concepts is here:
data/v2/open5e/core
The models considered core are:
- Ability (Strength)
- Alignment (Chaotic Good)
- CreatureType (Aberration)
- DamageType (Acid)
- Environment (Tundra)
- ItemCategory (Potion)
- ItemRarity (Uncommon)
- Language (Common)
- Size (Large)
- Skill (Survival)
- SpellSchool (Evocation)

This is missing a notable exception: Conditions. That is carved into a pair of follow up tickets, #791 and #793 . 

The above are core concepts, but in some cases descriptions will vary from document to document. This change ALSO introduces 5 Description models, each tied to one of the models above. This will allow for API clients to choose a description from a set of descriptions when necessary. The API does NOT make a choice of which description is correct.

Description data from SRD 2014 and SRD 2024 is added to all 5 of the models where descriptions are provided.

Another note: It is possible to add on to these. If a document adds a "Ginormous" size that's 100x100 feet, it is possible for the size itself to be completely defined within the scope of that document, the core document is NOT the only place for sizes to be added.

I do expect API clients to ALWAYS select the "core" document though, otherwise things may get strange.


augustjohnson:
These changes above resolve #790

augustjohnson:
Commits above remove the ConditionConcept from the code. This should resolve #791

augustjohnson:
#793 is addressed with the most recent commits.

calumbell (approved):
Hi, thanks for getting this change made. Code and and API responses are looking good to me. Tests are passing. Happy to approve this!

I initially had a few misgivings about how the structure of the API response, but following our discussions over Discord it makes sense why the API response is structured like this. Considering the intention behind the issue you are fixing, I don't think we necessarily need to be worrying about "how would this scale if we have 100s of definitions of what a potion is from different sources?". Your approach is very appropriate for our current data-set.

I was fully expecting that this PR would introduce some breaking changes on the front-end, for which I would have asked for some time to patch before we merged this PR to `staging`. But somehow everything appears to be functioning either exactly as it did before, or at least failing gracefully enough that I can't spot the errors. Nice work!

PR #795, opened 2025-09-23
Added `desc` field to CharacterClass model

### LINKED ISSUE #789: [FEATURE REQ.] Add nullable `"desc"` field to V2 Class model

While working on adding the SRD-2024 classes data to the Open5e dataset, I noticed that there was no field on the `CharacterClass` V2 model to capture this following copy:

Image

ibid for many other subclasses. 

As this data is essentially a description of the subclass, this seems like the perfect place to use the `desc` field. However, this field is not part of the `CharacterClass` model.

Add the `desc` (string) field to the CharacterClass model. This field should be nullable. Set its default value to `""` so that our existing data will work as is, although we'd like to eventually add the subclass descriptions to the other subclasses in the O5e dataset.

I considered using the `ClassFeature` model to encapsulate this information, but that felt really wonky to me. This subclass front-matter is a description and not a feature, but it there would be a misnomer and cause more confusion in the long run.

calumbell:
## Description

This PR adds the the `desc` field to the CharacterClass model, which will help support the ingestion of the 5.2.1 SRD subclass data which includes a brief piece of descriptive text in the front-matter.

This was achieved by having the CharacterClass model inherit from the HasDescription abstract model. The HasDescription model then had to be updated to support nullable description by the `blank=True` property to the `desc` field

## Related Issue

Close #789


augustjohnson (approved):
Overall fine, but this does have the consequence of allowing blanks in all of the other models. Probably not a problem, but we can no longer depend on it being enforced at the model layer.

Issue #801, opened 2025-09-29
Add descriptions for V2 subclasses


nstefanski:
Some subclasses in the V2 data are missing the "desc" field. This data *should* exist in the V1 data, so it just needs to be checked and ported over where needed.

calumbell:
Closed by #802

PR #802, opened 2025-09-29
Add V2 subclass descriptions and fix some V2 class features

nstefanski:
I'm starting to work through #801 and adding the "desc" field to the V2 subclasses.

In doing so, I noticed some classes did not have their class features ported over correctly from V1 to V2 so I am fixing and adding those as I find them.


calumbell (approved):
Thanks for your contribution! Everything looks good to me.

Tested by pointing a local deve version of the FE site at a Django dev server running this branch to confirm that all Markdown renders correctly (it does).

All Pytest cases passing. `quicksetup` runs without error, so I think we are safe from any API breaking JSON errors.

I'll get this merged to `staging`, thanks again!

PR #803, opened 2025-09-30
Add rest of V2 subclass descriptions and fix ToH V2 class features

nstefanski:
Also for issue #801, there were more fixes needed to class features than expected so I had to make another pull


calumbell (approved):
Ran a pretty thorough spot-check of the updated entries and everything looks good. Nice spot on the features that needed updating. All tests passing. Thanks for your work, i'll get it approved and merged : )

PR #807, opened 2025-10-03
Open5E content

nstefanski:
add School of Abjuring and Warding + Abjurer/Warder (Class of ‘24)


calumbell (changes requested):
Hi @nstefanski!

All the Markdown parses correctly, content looks good. Nice work!

Requesting a few changes because I think the title "Abjurer/Warder (Class of ‘24)" needs a bit of work:

- Lets drop the "(Class of '24)" in the parentesis (as much as I appreciate the pun). This information doesn't really belong in the display name (that it is a subclass for the 2024 Wizard is right there in the `"subclass_of"` field) and it also doesn't fit with established conventions across the dataset.

- Could we pick either Abjurer or Warder? Abjurer is preferred, for being a much more evocative term, but I don't think we can use it because that is the name of the non-CC WotC class. "Warder" as a straight agent noun feels a bit flat, and it also brings to mind a prison guard, in which case "Warden" feels nicer. My preference would be for something like "Wardwright". We always to go goofy high fantasy with something like "Wardweaver". I'll leave the call to you, just not "Abjurer/Warder" please : )

- Finally, update PKs accordingly

Apologies if this seems nit picky! This is more of copy-editing review than a code review, which have way less experience authoring

nstefanski:
I like the sound of "Wardweaver." However, I've also been trying to keep them close to the WotC names alphabetically so they'd be quick to find for those looking. I'd thought of "Abjurationist" for this reason even though it's clunky, and we could also do something fairly generic like "Abjuration School" or "Abjuration Wizard." Or, now that I think of it, when I was looking for a synonym for the Arcane Ward feature I settled on Arcane Aegis, so what about something like "Aegisweaver"?

I feel like "Abjuration Wizard" would be easiest from a user's point of view if they're already familiar with the 5e subclasses. "Aegisweaver" or "Wardweaver" or the like if we want something more poetic.

calumbell:
Thats a really good point about sort order! I actually quite like Abjurationist. It clearly points back to the Abjurist class, and has a whiff of verbosity (without being actually verbose) that feels appropriate for the 'book magic' class

nstefanski:
Changed to Abjurationist, all pks should be fixed now too

calumbell (approved):
Thanks for making those changes, everything looks good. Nice work! Will get this approved and merged.

I did spot a tiny bug in the parent of one of the `ClassFeatureItem`s. Went ahead and fixed it, hope you don't mind!

PR #808, opened 2025-10-04
Added BFRD Mechanist class to V2 Data

calumbell:
## Description

This PR was a stress-test to make sure that the beta version of the V2 Class model is flexible enough to fully represent the classes presented in the Black Flag Reference Document. I opted for the Mechanist, because no equivilent class exists in the current O5e dataset.

N.B. PR #799 includes an small update to the `ClassFeature` model. If PR #799 is merged before this PR, then `"feature_type"`s should be added to the JSON data in **this** PR.

If this PR is merged first, then the `"feature_types"` should be added to PR #799. I think it makses sense that #799 is merged first though


eepMoody:
Nice, this is an awesome stress test! Looks like things largely are working out?

eepMoody (approved):

PR #812, opened 2025-10-13
Expanded `ClassFeature` `"feature_type"` options

### LINKED ISSUE #811: [BUG] Eldritch Invocations/Metamagic Options are not easy to seperate from other ClassFeatures
## Description

Looking at the WotC 5.1 and 5.2 SRDs, we can see that the Warlock's list of Eldritch Invocations and the Sorcerer's list of Metamagic options are rendered *after* each of the class's features that are gained at X level under their own heading.

SRD 5.2.1, pp. 66 (2024 Sorcerer class data):

Image

SRD 5.1, pp 48 (2014 Warlock class data):

Image

We would like API users to be able to mimic this behaviour. However there isn't a simple way of achiving this with the current `ClassFeature` model.

## Desired Solution

In my opinion, the best and most idiomatic solution would be to add a new option for the `"feature_type"` field on the `ClassFeature` model. I am leaning toward a name like `"CLASS_OPTIONS"`, `"CLASS_OPTION_LIST"`, `"SUPPLEMENTRY_CLASS_OPTIONS"`, something like that.

Once the model has been updated so that this new options is added to the `"class_feature"`, the following `ClassFeatures` should be set to use it;

- Warlock Eldritch Invocations (`srd-2014` and `srd-2024`)
- Sorcerer Metamagic Options (`srd-2024`)

calumbell:
## Description
This PR expands the `ClassFeature` model so that the `"feature_type"` now includes the option `"CLASS_FEATURE_OPTION_LIST"`. This is to facilitate easier seperationg of lists of class options that typically go after the features per level list (Warlock Eldritch Invocations, Sorcerer Metamagic options, etc.)

The V2 was also updated so that the SRD-2014/2024 Invocation list and the SRD 2024 Sorcerer Metamagic list have their `"feature_type"` field set to this new value.

## Related Issue

Closes #811 

## How was this tested

- `pytest` (all tests passing)


augustjohnson (approved):
Reasonable changes to support these use cases.

PR #813, opened 2025-10-17
Open5e 2024

nstefanski:
I noticed the new 2024 Abjurationist subclass wasn't showing up on the beta site (https://beta.open5e.com/classes/srd-2024_wizard), and it occurred to me that open5e content was only set up for 2014. I added a new open5e-2024 and moved the Abjurationist content to it. Future 2024 subclasses can go here.


augustjohnson (approved):
Remapping makes sense.

PR #814, opened 2025-10-19
Ingest Spells That Don't Suck

somanyrobots:
Pulling in the spells from the [Spells That Don't Suck](https://www.somanyrobots.com/s/Spells-That-Dont-Suck-compressed.pdf) project. The whole set is CC-BY (also, I'm one of the rightsholders). This is only the first draft (I still need to validate, test, and do more error-checking), but I wanted to get the PR up as a heads-up on the effort. 

These were generated from the project's foundry module via a ruby script; happy to share that script up once all the kinks are ironed out if it would be of use (also, I'll probably do another spells import with it).


somanyrobots:
Okay, that should have everything working in here. Later this week I'll try to PR up a few issues with existing spells that I noticed in the course of this.

somanyrobots:
guh, just realized none of the classes imported properly. will get a fix up for that

somanyrobots:
okay, did a big pass cleaning everything up - corrected a bunch of underlying data mismatches in the foundry module (mostly stuff that didn't have `target` values specified). had to do a big detour because in the process of updating the foundry module, foundry auto-updated a bunch of spells to use the `activity` pattern introduced in the last release, so I had to write a parser for that. Should be good to review now!

augustjohnson (approved):

PR #842, opened 2025-12-25
Fixed parsing reaction conditions without parentheses

blakewatson:
Many spells like Shield and Feather Fall which have a reaction were missing the conditions of the reaction. I wasn't smart enough to write one regex to rule them all. What I ended up doing was checking if the first regex failed and if so searching again without looking for parentheses and also allowing for an optional comma.

This could probably use some sort of fancy diff on the resulting JSON file to make sure nothing got screwed up. But I did confirm that some spells that were missing the reaction condition before now have them.


eepMoody (approved):
Thanks for this, nice catch!

PR #858, opened 2026-01-12
Ingest KibblesTasty spells

somanyrobots:
Adding KibblesTasty spells from KCCC and KCLL; had to reopen the PR because it wouldn't rebase cleanly on all the staging changes I hadn't copied over.

All these spells are CC-BY-licensed (even though they're in paid books) and have express permission from the creator to import.


somanyrobots:
This is now passing tests, but still not producing totally correct outputs; I'll mark as ready once I'm done QA'ing and regenerating.

somanyrobots:
okay, I think this is good to go - I reviewed all the spellcasting options, tweaked the scripts, and fixed a dozen or so by hand that weren't worth the trouble to adjust the scripts for.

calumbell (changes requested):
Hey @somanyrobots, thanks for getting this PR submitted. Obviously kind of difficult to QC all the spells included in here, but those I looked at on my cursory spot check all looked good to me.

These are a couple of changes it would be good to resolve before we get this merged.

1. I see that you have added stubs for new classes, but all the class feature data for these is missing. Could we either:
- (A) Remove the references to these classes from the Spell `"classes"` field until they are implemented.
- (B) Add in the missing `ClassFeature` and `ClassFeatureItem` data required to complete these classes (which is a fair bit of work, i'd be happy to offer a hand)

2. The `key` fields for the new `Document` data look a bit strange. The schema for appreviated the title appears to be mixed. Lets pick one scheme and run with it:
- (A) `kibbles-krd`, `kibbles-ccc`, `kibbles-cll`
- (B) `krd`, `kcc`, `kll`
- (C) `kibbles-krd`, `kibbles-cc`, `kibbles-ll`

Thanks again for your contribution! Let me know if you have any questions

somanyrobots:
alright, made the easy change. I'll take a crack at the hard change - parsing the class data out of the foundry module is actually probably easier than the spells were, I won't need the same regex forest.

PR #884, opened 2026-02-26
Add ability to filter monsters by environment

stevennoto:
This PR adds a filter to monsters on the field environments_json. It enabled filtering monsters by their environment.

Example API call now possible: `/v1/monsters/?limit=50&type=Dragon&environments=Mountain`

The filter uses simple substring matching, since the environments json data includes simple strings with some inconsistencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ pipenv run python manage.py spectacular --color --file openapi-schema.yml` to bu
 ```
 
 # Contributing
-See [contribution guide](CONTRIBUTING.md).
+See [contribution guide](.github/CONTRIBUTING.md).
 # Tests
 
 Tests are located in the `api/tests` directory. These should be run before pushing new changes to the main repository. These tests require that the api is [running](##run) at `http://localhost:8000`.
--- a/api/filters.py
+++ b/api/filters.py
@@ -76,6 +76,10 @@ class SpellListFilter(CommonFilterSet):
 
 
 class MonsterFilter(CommonFilterSet):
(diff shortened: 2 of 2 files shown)


calumbell (approved):
Hi @stevennoto, thanks for your contribution! I have tested this out on my local Django development server and all appears to be functioning as intended.

And thanks for fixing that deadlink in the README! I shudder to think how long that has been there for...

I'll approve and get this merged to `staging`

stevennoto:
Thank you @calumbell for the fast review and merge, that's awesome!

Do you know if any changes to the website would be needed to expose the new filter at https://open5e.com/monsters ? If so, let me know, and I can take a look. Thanks!

calumbell:
Happy to help @stevennoto, thanks for contributing!

On the topic of changes we'd need to see to expose the filter on the front-end, there is a little complexity to explain:

There are currently two versions of the Open5e API: V1 and V2. The current website uses V1, but we have been working on migrating both the front- and back-end to V2 for a while now. We are at the finishing touches, polishing phases. Release is Very Soon (TM). The most up-to-date version of the Open5e site (which can be found here: https://beta.open5e.com/) uses the new V2 API exclusively. This will overwhelmingly likely be our next cut to production.

Your PR added the new filtering options to the V1 version of the API, which we plan to maintain for legacy purposes, but we won't be using on our live site anymore. So, to answer your question, to support filtering monsters by environment on the website we'll need to update the V2 filters (likely a very small change).

Feel free to open an issue, take a stab at it yourself if you have the bandwidth!

Issue #885, opened 2026-02-26
[BUG] Leading slashes in `v2/image` `"file_url"`


calumbell:
## Bug description

While working on #881, after installing the Django Debug Toolbar, I noticed that one of our tests was failing. The URL being tested was `/v2/conditions`. The full stack trace can be viewed below.

Investigating the `/v2/conditions` endpoint and was greated with an error page that read:

```
SuspiciousFileOperation at /v2/images/

The joined path (/img/object_icons/elderberry-inn-icons/conditions/blinded.svg) is located
outside of the base path component (/Users/calum/Documents/code/open5e-api/static)
```

Which pointed me toward the `/v2/images` endpoint where I experienced the same error.

Digging into this error I believe that it is being caused by the leading slash in the `"file_path"` in our V2 Image data. This is used to specify an absolute path rather, which I believe is causing the Django development server to throw a `SuspiciousFileOperation` error when Django tries to reach 'up' belong to repository's root director to the computer's root. **The fix here is to remove the leading '/' on the `"file_url"` fields**.

**The error does not appear if DEBUG mode is off**. So this shouldn't cause any issues on the live API, only during local testing.

I can only assume that this error has emerged because the DDT is somehow interferring with how the other middleware (WhiteNoise?) is handling our static files, specifically on sanitising those leading slashes.

So while there is a deeper issue to unpack here about dev/prod parity, the leading slash in the `"file_path"` field of our image data is nonetheless a data error and ought to be fixed.

## Code 

<details>
<summary>Pytest log</summary>

```
___________________________________________________________________ TestObjects.test_condition_example ___________________________________________________________________

self = <Response [400]>, kwargs = {}

    def json(self, **kwargs):
        r"""Decodes the JSON response body (if any) as a Python object.
    
        This may return a dictionary, list, etc. depending on what is in the response.
    
        :param \*\*kwargs: Optional arguments that ``json.loads`` takes.
        :raises requests.exceptions.JSONDecodeError: If the response body does not
            contain valid json.
        """
    
        if not self.encoding and self.content and len(self.content) > 3:
            # No encoding set. JSON RFC 4627 section 3 states we should expect
            # UTF-8, -16 or -32. Detect which one to use; If the detection or
            # decoding fails, fall back to `self.text` (using charset_normalizer to make
            # a best guess).
            encoding = guess_json_utf(self.content)
            if encoding is not None:
                try:
                    return complexjson.loads(self.content.decode(encoding), **kwargs)
[...]

PR #887, opened 2026-02-27
Install Django Debug Toolbar

### LINKED ISSUE #881: Add Django Debug Toolbar as a dev dependency
On several occasions I have been able to to diagnose N+1 errors on the API using the [Django Debug Toolbar](https://django-debug-toolbar.readthedocs.io/en/latest/index.html). It has been a really helpful tool and I think it would be well worth adding 

Install Django Debug Toolbar as a development dependency

So far, I have been installing/uninstalling DDT on my local fork whenever I need it. There is a lot of friction involved here, and I thinking developers would be better served with having it already configured.

**Additional context**
Initial tests have shown that the toolbar is disengaged when the `DEBUG` settings flag is set to off. Additional care should be taken to ensure that DDT is completely disengaged during production to avoid leaking any data.

calumbell:
## Description
This PR adds the Django Debug Toolbar to Open5e as a development dependency. It is only configured to appear when the `"DEBUG"` flag in `server/settings.py` is set to True.

Screenshot 2026-02-27 at 09 05 19

I have used the DDT on my local fork for some time now, and it has been really helpful for resolving N+1 errors that occasionally crop up.

## Related Issues
Closes #881 

## How was this tested

N.B. local tests are failing on this branch due to issue #885. It would be preferred if PR #886 (which fixes the bug) were merged first to avoid this causing any confusion.


calumbell:
Closing and re-opening to re-run validation

calumbell:
Hmm, struggling to get debug mode and dev dependencies playing nicely today. Will close for now and have a think

PR #899, opened 2026-03-20
[BUGFIX] `GameContentSerializer` now inherits from `ModelSerializer`

### LINKED ISSUE #896: [BUG / PERFORMANCE] slow Sumary serializers generating unnecessary URLs
## Bug description

While investigating slow performance on the `/v2/creatures` endpoint using Python's `cProfile` library, I discovered that the calls to the `url()` function were taking up a significant amount of server resources. When requesting a page of 50 creatures, 600+ calls to `url()`, taking up close to 40% of the server response time.

After some investigation, I believe that I have located the culprit:

```
class GameContentSerializer(serializers.HyperlinkedModelSerializer): 
  ...
```

The `GameContentSerializer` is inherited but most/all of our other serializers and handles a lot of the API's features. This serializer in turn inherits from the DRF's `HyperlinkedModelSerializer` which extends the `ModelSerializer` which functionality to support easily generating URLs.

What I believe is happening on this endpoint is that Serializers that really don't need require a URL field, I am thinking of the Sumary serializers (`DocumentSumarySerializer`, `SizeSumarySerializer`, etc.), which were designed to serializer FKs on model views and designed to be as streamlined as possible, are generating their own URLs *even* when the URL is not being returned by the View.

So, hitting the `/v2/creatures`, every FK (document, document__publisher, size, creature_type, etc.) would be generating its own URL, which has become very expensive.

## Proposed Solution

The best way to eliminate the unnecessary URL generation would be to:

1. Refactor GameContentSerializer to inherit from `serializer.ModelSerializer` instead of `serializers.HyperlinkedModelSerializer` to stop the automatic generation of URLs.
2. [OPTIONAL] Explicitly add the `"url"` field back onto serializers where it makes sense using a `HyperlinkedIdentityField`.
3. Remove `"url"` from serializer fields inheriting from `GameContentSerializer`
4. Update tests to reflect new shape of API response

calumbell:
## Description
This PR fixes an issue where unnecessary URLs were being generated for some of our find many endpoints (ie. fetching 50 Creatures caused about 600 calls to `url()` according to `cProfile`) taking up a significant amount of server time.

The issue was there `GameContentSerializer`, the abstract class inherited by the vast majority of our Serializers, inherited from the `HyperlinkedModelSerializer`, which *always* generates a URL when it is called, whether or not the serializer even returns a URL, or if it makes sense for it to!

The fix involved the following steps:
1. Replace `HyperlinkedModelSerializer` with `ModelSerializer` in `GameContentSerializer`'s inheritance chain
2. Remove `"url"` from the `"fields"` array in serializers that inherit from `GameContentSerializer`.
3. Update our test suite to account for the removal of the `"url"` field from many serializers

## Related Issue
Closes #896

## How was this tested?
- `pytest` (all tests passing, once cases updated to reflect new shape of API response)


calumbell:
Closing and re-opening to re-run PR validation

eepMoody (approved):

PR #924, opened 2026-05-19
[BUGFIX] Fixed crashes on Search

### LINKED ISSUE #922: [BUG] Search is crashing (500 res) when certain common queries are searched
## Bug description
When searching certain terms using the `/v2/search` endpoint, the server throws a 500 (Internal Service Error). On the front-end, this cause the website to crash.

Investigating the `staging` branch on local, I caught the following information from the stack trace in the terminal when attempting to search the query **"dragon"**:

```
  File "/xyz/Documents/code/open5e-api/search/serializers.py", line 67, in get_object
    return result_detail.search_result_extra_fields()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/xyz/Documents/code/open5e-api/api_v2/models/document.py", line 149, in search_result_extra_fields
    "school":self.school.key,
             ^^^^^^^^^^^
AttributeError: 'Item' object has no attribute 'school'
[15/May/2026 21:34:02] "GET /v2/search/?limit=5000&document__key__in=no-sources&schema=v2&query=dragon HTTP/1.1" 500 22877

```

It looks like the search serializers are attempting to the index into the `'school'` field of the `Item` serializer, which doesn't have that field (it is a `Spell` field).

## Steps to Reproduce

Visit the Open5e website and use the search bar to search for any of the following terms:
- Dragon
- Sword
- Shield
- Ring
- Strength

And you will hit the error page (500: Internal Service Error)

However, the bug does _not_ occur when the following terms are searched for:
- Goblin
- Fireball
- Devil
- Acolyte
- Fighter

calumbell:
## Description

This PR fixes a bug that was causing the v2 Search to crash when certain terms were searched for (see issue #922).

The bug appears to be in the code that adds additional contextually relevant fields to search results. adding a spell's school and level when the search returns a spell, for example.

The `Document` model's `search_result_extra_fields` method instructed the serializer to add the `"school"` field to any search result that inherits from the `FromDocument` model.  I can only assume that this bug only triggered on some search terms and not others was because in many cases the `search_result_extra_fields` is overloaded in the Class that inherited from the `FromDocument` abstract model, obscuring the bug. A quick peak at the error messages/stack trace reveals that it was the `Item` model which was giving the API issues, and investigating the Item model it looks like this method is not defined there. Surprise surprise.

This would normally be the type of bug that `pytest` would catch, but it looks like the Search is a blind-spot in our testing suite. Would be well worth adding this in!

## Related Issue

Closes #922 

## How was this tested?
- `pytest` (all tests green)
- All CI tests passing

### DIFF
--- a/api_v2/models/document.py
+++ b/api_v2/models/document.py
@@ -145,9 +145,7 @@ class FromDocument(models.Model):
         return reverse(self.__name__, kwargs={"pk": self.pk})
 
     def search_result_extra_fields(self):
-        return {
-            "school":self.school.key,
-        }
+        return
 
     class Meta:
         abstract = True


augustjohnson (approved):
Makes sense, no concerns.

PR #929, opened 2026-05-19
Made cache clear dependent on the other workflow.

### LINKED ISSUE #815: [BUG] Clear Cloudflare cache completes before the push does.
## Bug description
The Clear Cloudflare Cache action ends up completing before the build and deploy action does, meaning the cache can get into some funky states.

## Steps to Reproduce
Merge a PR to either staging or main. This will kick off the Clear Cloudflare Cache action, and the Build and Deploy action.

Often (always?) the Clear Cloudflare Cache action will complete before the build and deploy action will.

augustjohnson:
Changed the trigger on this one to be dependent on the Build and deploy action, only when completed. Then the cache can be cleared. Should be improved.

### DIFF
--- a/.github/workflows/clear_cache.yml
+++ b/.github/workflows/clear_cache.yml
@@ -1,7 +1,10 @@
 name: Clear Cloudflare Cache After Deployments
 
 on:
-  push:
+  workflow_run:
+    workflows: ["Build and deploy"]
+    types:
+      - completed
     branches:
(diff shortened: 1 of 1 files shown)


calumbell (approved):
Not sure if there is an easy way to test out this workflow without merging, but I went over the Github Actions documentation to make sure everything here makes sense, and it does!

PR #932, opened 2026-05-22
718 technical debt   align model field names and serializer names

### LINKED ISSUE #718: Technical Debt - Align model field names and serializer names
With recent changes to the creature response in #717 there are some fields that are mis-matched. Ideally they should align so that data entry is good to go, but this takes some coordination and some data massaging to ensure it works correctly.

There may be fields on other models that require this refresh as well. This issue includes researching which ones are mismatched.

augustjohnson:
This boiled down to the chosen display side of CreatureAction being different than the model stuff. All changes were made to align those.

All other similar scenarios had solid reasons to NOT refactor.
- CreatureLanguageSerializer as_string is nested on purpose.
- resistances_and_immunities is grouped, doesn't align with backend stuff.
- General SerializerMethodFields - have logic or combinations in general, not good candidates for readjustment.
- experience_points - another computed field.

No tests changed values, all tests remained passing.


calumbell (approved):
Looks good on local with the front-end pointed at it. Changes to the actual logic are small and sensible in spite of the large number of lines changed

PR #945, opened 2026-05-31
926 migrate beta deploy to use ghcr instead of docker hub 2

### LINKED ISSUE #926: Migrate beta deploy to use GHCR instead of docker hub.
The current build_and_deploy.yml workflow pushes to Docker Hub, which requires maintaining separate creds (DOCKERHUB_USERNAME, DOCKERHUB_TOKEN) as repo secrets. The image tag uses staging, which doesn't match the environment name (beta / api-beta). If multiple PRs merge to staging in quick succession, multiple builds and deploys run simultaneously, racing against each other on the DigitalOcean deploy API.

- Use GHCR instead of Docker Hub (as the default at least, potentially mirror on docker hub down the road).
- Add a concurrency protection to this to ensure reliable builds.
- Add filters for docs-only PRs.
- Update the DO deploy to use GHCR instead.

- *Stay on docker hub* - works but spans +1 services, and includes an individual's tag (eepmoody) as part of the container repo.

**Additional context**
Post issue resolution, PRODUCTION will still build from docker hub. That's future work. This issue is scoped to be only beta deploys.

augustjohnson:
This PR updates the CI/CD pipeline to publish the staging/beta image to GitHub Container Registry (GHCR) in addition to Docker Hub, and adds two workflow quality-of-life improvements.

1. The staging branch now pushes to both eepmoody/open5e-api:staging (Docker Hub) and ghcr.io/open5e/open5e-api:beta (GHCR). This will be a longer-term goal, but for now both locations are targets.
2. Added paths-ignore to skip CI on doc/script-only changes, and added a concurrency group so in-progress deploys to the same branch are cancelled when a newer push arrives.
3. .dockerignore: This branch adds a root-level .dockerignore to prevent .git, caches, local env files, markdown docs, and other non-essential files from being sent to the Docker build context. data/, source dirs, pyproject.toml, uv.lock, and manage.py are intentionally kept as they are required by the multi-stage build.


augustjohnson:
This is ready for review!
The bonus changes here at the end had a few non-functional changes. The dockerfile should be significantly smaller now (no /data directory), and quicksetup should run faster due to a couple of tweaks without workflow impacts.

PR #956, opened 2026-06-09
[DX] Added `django-debug-toolbar` as a dev dependency

calumbell:
## Description

This PR adds the Django Debug Toolbar to our list of development dependencies. This package has been extremely helpful in identifying N+1 errors on the API and is well worth having installed as default.

## How was this tested
- (Local) Changing `OPEN5E_DEBUG` environment variable to False to confirm that the toolbar *didn't* run outside of debug mode.
- (Local) Run `pytest` to make sure the toolbar didn't interfere with it.

Pending to see whether the the toolbar is able to get out of the way correctly during CI tests. Standby.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,7 @@ dev = [
     "pytest-django",
     "approvaltests",
     "pytest-approvaltests",
+    "django-debug-toolbar>=6.3.0",
 ]
 
 [tool.uv]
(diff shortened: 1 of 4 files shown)


calumbell:
Alright, looks like CI is failing. After a brief investigation the problems appears to be on the Docker image build step:

```
#10 ERROR: process "/bin/sh -c uv sync --frozen --no-dev --no-install-project" did not complete successfully: exit code: 2
------
 > [builder 4/7] RUN uv sync --frozen --no-dev --no-install-project:
0.184 error: Required uv version `>=0.10` does not match the running version `0.9.30`
------
```

Looks like some kind of `uv` version mismatch. Will investigate further later this week.

